PR #197, opened 2021-06-14
Use URLSearchParams

B0pol:


FireMasterK:
Nice!

Although, at this point, wouldn't it make sense to construct a `URL` object and modify it with `url.searchParams.set(...)`?

FireMasterK:
I also just thought of a way to take the idea further - creating a mixin which would further simplify it. Where the parameters would just be th URL and object which would be converted to query.

I will try implementing this tomorrow.

FireMasterK:
I've implemented my approach in #198, this approach also makes the parameters handled like how they are in `requests` - the python library.

My approach also reduces a significant amount of code lines, which is why I will close this in favor of that.

PR #198, opened 2021-06-15
Handle parameters more gracefully with the mixin.

FireMasterK:
Alternative take on #197


FireMasterK:
Merging as nothing seems to be broken.

PR #211, opened 2021-06-21
Add links to public chat rooms

B0pol:


FireMasterK:
This patch seems to put them in the same line when rendered. It would be nice to have them on different lines. (The previous approach used bullet points for this)

PR #219, opened 2021-06-22
Add light theme.

### LINKED ISSUE #130: Light Theme
I understand the importance of having a dark theme, and they many people prefer it, however not everyone does, and many (including me) prefer having a light theme (as long as it isn't too blinding).

FireMasterK:
Closes #130


moriel5:
I had just tested, and the results were not very good (pictures uploaded to Issue #130), so that would certainly need to be improved (great job, however).
It would be nice if, once the theme is fully functional, that also the settings and channel pages were also themed with the light them.

PR #272, opened 2021-07-18
Adjustments to WatchVideo.vue to be more responsive and more similar to Youtube's UI

garrettjoecox:
Comments are now displayed below the related videos, similar to youtube UI. Related videos and comments now take full width for smaller screen.

Before:
Screen Shot 2021-07-18 at 10 41 29 AM
After:
Screen Shot 2021-07-18 at 10 42 01 AM

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -63,7 +63,7 @@
         <hr />
 
         <div uk-grid>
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thanks! This seems better than the current solution, so I'll take this!

I would ideally like the suggestions to be minimizable, like on the mobile website, but that can be done later!

PR #273, opened 2021-07-18
Give the video meta data section a bit of breathing room

garrettjoecox:
Also pulled the subscribe button over from the channel vue, though I'm not super happy with using `.split('/')` on the uploaderUrl to get the ID...

And lastly, changed the "-" / "+" to something more descriptive, since there's not much conflicting for that space anyway.

Before / After
Screen Shot 2021-07-18 at 12 29 51 PM


FireMasterK:
Hey, thanks for this!

For the linter, you can use Vetur, a VSCode extension, with Pretier in the future, it should follow the project's style then :)

PR #291, opened 2021-07-20
Show video duration in thumbnail (#269)

### LINKED ISSUE #269: Show video duration in the thumbnail
Can you put a video's duration alongside its thumbnail in places from where you'd select to watch videos from (search results page, Trending page, etc.)? It's a basic thing I expect to be able to compare while choosing a video to play.

carlssonemil:
Moved the video duration from the video metadata section to a label in the bottom right corner of the thumbnail. Fixes #269.

### DIFF
--- a/src/components/FeedPage.vue
+++ b/src/components/FeedPage.vue
@@ -18,7 +18,14 @@
         >
             <div class="uk-text-secondary" :style="[{ background: backgroundColor }]">
                 <router-link class="uk-text-emphasis" v-bind:to="'/watch?v=' + video.id">
-                    <img style="width: 100%" v-bind:src="video.thumbnail" alt="thumbnail" loading="lazy" />
+                    <div class="uk-position-relative">
+                        <img style="width: 100%" v-bind:src="video.thumbnail" alt="thumbnail" loading="lazy" />
+                        <span 
+                            v-if="video.duration"
+                            class="uk-label uk-border-rounded uk-position-absolute video-duration" 
+                            style="bottom: 5px; right: 5px; background: rgba(0, 0, 0, .75); color: white; padding: 0 5px;"
+                            >{{ timeFormat(video.duration) }}</span>
+                    </div>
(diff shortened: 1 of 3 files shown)


FireMasterK:
Sweet! Do you think it would be possible to invert the backgrounds? - Have the background black, and have the text white, to keep it similar to how YouTube is

Your PR:
image

YouTube:
image

carlssonemil:
Sure thing! I tried adding the `.ui-dark` class but it's overridden from higher up in the hierarchy by `.ui-light`. Will look at it tomorrow, time for bed now.

Also, a question, why isn't the component [VideoItem.vue](https://github.com/TeamPiped/Piped/blob/master/src/components/VideoItem.vue) used in all the views where a video is visible? It would make it easier to make changes. I tried using it for the SearchResults but the model passed from the API is not the same, for example in the trending page `title` is used but in the search results it's `name`.

FireMasterK:
> Also, a question, why isn't the component VideoItem.vue used in all the views where a video is visible?

This is only because search results can also be playlists and channel, some work would be needed if necessary to be adapted by the feed/search results.

> the model passed from the API is not the same, for example in the trending page title is used but in the search results it's name

This would be a good time to unify all the field names! I'll work on this tomorrow.

carlssonemil:
> This is only because search results can also be playlists and channel, some work would be needed if necessary to be adapted by the feed/search results.

Oh, right! That makes sense. One way to fix that is to maybe add a `type` property to the search result object, defining if the item is a video, channel or playlist and then use `v-if` to show the correct component?

carlssonemil:
I changed to black background and slightly reduced the horizontal padding since it wasn't really necessary. See picture below.

image

And added the changes to the [FeedPage.vue](https://github.com/TeamPiped/Piped/blob/master/src/components/FeedPage.vue) aswell.

FireMasterK:
Awesome, thank you!

PR #292, opened 2021-07-20
Vue doesnt change execute `mounted()` on vue route change, so it's better we use a watcher to set page titles.

IvRRimum:
https://stackoverflow.com/questions/57699473/vuejs-hook-mounted-is-not-called-when-page-is-re-open-by-vue-router

Vue doesnt change execute `mounted()` on vue route change, so it's better we use a watcher to set page titles.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -30,6 +30,22 @@ export default {
     components: {
         Navigation,
     },
+    watch: {
+        $route: {
+            immediate: true,
+            handler(to) {
+                switch(to.path) {
+                    case "/feed":
+                        document.title = "Feed - Piped";
+                        break;
+                    case "/":
+                        document.title = "Trending - Piped";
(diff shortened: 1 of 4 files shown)


FireMasterK:
Actually, Vue does call mounted on route change!

The problem here is different, a week or so back, I introduced keep-alive for components, which does not call mounted when you open a cached component.

The way I would like to fix this is by moving the `document.title = ` calls to the activated method, since that is what is called when a component is resumed from its cached state.

IvRRimum:
Makes sense, but having a watcher also isn't technically wrong. I can only imagine anomaly in title setting when video/playlist loads and for all other pages we could have a single location where we set the title.

IvRRimum:
Oh shoot, i made a pull request with the master branch, let me close this.

PR #297, opened 2021-07-21
Fix comment formatting (#267)

### LINKED ISSUE #267: Fix comment formatting
First of all, just wanted to say congrats on the project. Loads way faster than Invidious!

The one thing that stood out was how the comment user avatar and comment were on separate rows. I'm currently using the following user CSS so the comment avatar is floated to the left, followed by the comment, but feel free to adapt it to work with Piped:

```CSS
div.uk-align-left div[align="left"] {float:left; width:90px; word-wrap:break-word;}

div.uk-align-left p {float:left; margin-left:1em; margin-top:0; width:calc(100% - 1em - 90px)}

div.uk-align-left p + div {clear:both;}
```

carlssonemil:
I fixed the comment formatting from this:

To this:

I also added a [Comment](https://github.com/carlssonemil/Piped/blob/fix-comment-formatting/src/components/Comment.vue) component to minimize the HTML code in [WatchVideo.vue](https://github.com/TeamPiped/Piped/blob/master/src/components/WatchVideo.vue). This would fix #267.

### DIFF
--- a/src/components/Comment.vue
+++ b/src/components/Comment.vue
@@ -1,0 +1,47 @@
+<template>
+  <div class="comment uk-flex">
+    <img
+        :src="comment.thumbnail"
+        class="comment-avatar uk-border-circle uk-margin-right"
+        style="width: 48px; height: 48px;"
+        loading="lazy"
+        alt="Avatar"
+    />
+
+    <div class="comment-content">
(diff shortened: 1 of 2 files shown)


FireMasterK:
Thank you, this is spot on!

PR #303, opened 2021-07-22
Sort countries in the Country Selection dropdown by name

venn-xx:
Sort countries by name, not code. Country lists are typically read in alphabetical order of name, not code.


FireMasterK:
Thank you!

PR #304, opened 2021-07-22
feat(hyperlinking): hyperlinking URLs in the channel descriptions

dotamir:
Fix for #7.
Now with this fix, URLs on channel descriptions are clickable as hyperlinks.

**Note**: Regex for URLs detection is not quite a complete regex. As you know there are complex and more completed regex out there. If you prefer to use a more confident regex, let me know.

Example channel: [here](https://piped.kavin.rocks/channel/UCKlOmM_eB0nzTNiDFZibSSA)

Comments are welcome.

### DIFF
--- a/src/components/Channel.vue
+++ b/src/components/Channel.vue
@@ -4,7 +4,7 @@
     <div v-if="channel" v-show="!channel.error">
         <h1 class="uk-text-center"><img height="48" width="48" v-bind:src="channel.avatarUrl" />{{ channel.name }}</h1>
         <img v-if="channel.bannerUrl" v-bind:src="channel.bannerUrl" style="width: 100%" loading="lazy" />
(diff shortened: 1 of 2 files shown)


FireMasterK:
Hey! This would introduce a severe XSS vulnerability!

I will take a look at this tomorrow.

dotamir:
Hey. You're right. So we need to use purify html here?

FireMasterK:
Yes, you're right, there's a mixin for it too. 
v-html calls are unsafe, hence why we use dompurify.

dotamir:
@FireMasterK Just pushed a new commit for the issue you mentioned.

FireMasterK:
Finally got a chance to test it and it works perfectly as expected, thanks!

PR #318, opened 2021-07-26
Fix overflowing titles

### LINKED ISSUE #306: Videos with long names overflow
When scrolling through channels that have videos with long names (see [this VSauce video](https://piped.kavin.rocks/watch?v=Pxb5lSPLy9c) as an example or [their Channel page](https://piped.kavin.rocks/channel/UC6nSFpj9HTCZ5t-N3Rm3-HA)) the names of those videos overflow past the view of the screen, which is especially apparent on mobile devices.
image

carlssonemil:
Added class `uk-text-break` to video title to stop it from overflowing. See pictures.

Before:
before

After:
after

This fixes #306.

### DIFF
--- a/src/components/VideoItem.vue
+++ b/src/components/VideoItem.vue
@@ -17,7 +17,7 @@
                     >{{ timeFormat(video.duration) }}</span
                 >
             </div>
-            <p>{{ video.title }}</p>
+            <p class="uk-text-break">{{ video.title }}</p>
         </router-link>
 
         <div :class="{ 'uk-align-left': !(video.views >= 0 || video.uploadedDate) }">


ajkblue:
I'm not an expert in coding Vue, but looking at that commit you made it looks like adding the same class of "uk-text-break" to [line 13 of WatchVideo.vue](https://github.com/TeamPiped/Piped/blob/9068564500b3533a848a27b5790be55c10155c0c/src/components/WatchVideo.vue#L13) also fixes the overflow issue on the video play page itself.

FireMasterK:
Thank you!

carlssonemil:
> I'm not an expert in coding Vue, but looking at that commit you made it looks like adding the same class of "uk-text-break" to [line 13 of WatchVideo.vue](https://github.com/TeamPiped/Piped/blob/9068564500b3533a848a27b5790be55c10155c0c/src/components/WatchVideo.vue#L13) also fixes the overflow issue on the video play page itself.

You're rioht! I made a new PR (#319) fixing this.

PR #323, opened 2021-07-29
fix import of Google Takeout CSV

avo7:
the file that Google has generated for me had the field `Channel ID` (capital "D") in the header.

after this small change, it worked for me.

i don't know if `Channel Id` is also valid in some cases, or not, so I left both the possibilities


FireMasterK:
Thanks!

Looks like Google is actively changing the takeout page, maybe this can be further improved in the future.

PR #335, opened 2021-08-01
10 second scroll on double tap player

IvRRimum:
This comes at the cost of double tap to fullscreen, i guess we could override the the shaka player logic and make and edge case for mobile devices, like here:
https://github.com/google/shaka-player/commit/cf5d893d6ff3dc43da89131570debc30348a1437

### DIFF
--- a/src/components/Player.vue
+++ b/src/components/Player.vue
@@ -241,6 +241,27 @@ export default {
         },
     },
     activated() {
+        const videoEl = this.$refs.videoEl;
+        var lastClick = new Date();
+        document.addEventListener("click", function(event) {
+            var t = event.target;
+            while (t && t !== this) {
+                if (t.matches(".shaka-mobile.shaka-video-container")) {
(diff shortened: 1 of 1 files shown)


IvRRimum:
But then again, double tap to fullscreen is a lot less used than double tap to 10 sec forward/backward, so i guess it works for now

FireMasterK:
I really want to use https://github.com/google/shaka-player/pull/3373 instead.

IvRRimum:
Like integrate this feature within shaka player api? Or do you want to have this feature integrated into the shaka player repo?

FireMasterK:
I'm interested in the PR in shaka-player, that adds native support for this. I want to avoid additional code here if possible.

IvRRimum:
Ohh, well it depends on them if they want to merge that feature. I will have time next weekend, might be able to do that.

Bnyro:
https://github.com/shaka-project/shaka-player/pull/5943

PR #362, opened 2021-08-16
Add j, k, l keybindings to Player

jim3692:
Based on official YouTube keybinds:
- J goes back by 15 seconds
- L goes forward by 15 secongs
- K is play/pause, like Space


FireMasterK:
Perfectly done, thanks!

PR #368, opened 2021-08-21
Changed LBRY references to Odysee references.

ProgrammerWhoPrograms:
Odysee is not LBRY. LBRY is a blockchain and P2P network for hosting
content without censorship, Odysee is a centralised site that serves
content from LBRY.
See lbry://@lbry:3f/theendoflbrytv:0 or https://odysee.com/@lbry:3f/theendoflbrytv:0 for more details.

Another thing that might be widespread in the future is redirecting 'lbry://' links to a native LBRY application. If that happens I think it would be worth adding a real 'Watch on LBRY' button with such a link. I may do that in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ By using Piped, you can freely watch and listen to content freely without the fe
 -   [x] Login
 -   [x] Feeds
 -   [x] Integration with SponsorBlock
--   [x] Integration with LBRY
+-   [x] Integration with Odysee
 -   [x] 4K support
 -   [x] No connections to Google's servers
 -   [x] Playing just audio
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -51,7 +51,7 @@
                     class="uk-margin-small-left uk-button uk-button-small"
                     style="background: #222"
                 >
-                    <b>Watch on LBRY</b>
+                    <b>Watch on Odysee</b>
                 </a>
             </div>
 


FireMasterK:
> Odysee is not LBRY.

Yes, but Odysee serves content from LBRY.

> Odysee is a centralised site that serves content from LBRY.

Odysee just has a public `lbrynet` node that is used to fetch data? It's like calling IPFS centralized just because you use a public gateway like `ipfs.io`.

Calls on Odysee:
image
Calls on LBRY Desktop:
image


I'm sorry, but I see Odysee as a way to watch content on LBRY by using a node hosted by someone else so that you don't need to. The LBRY desktop application uses the same API calls, but on a local node.

Certain things like comments, search, etc are centralized on both LBRY desktop and Odysee. I'm sorry to break it to you, but LBRY desktop isn't decentralized in the way you thought.

Here's some requests made on LBRY desktop:
image
image


> Another thing that might be widespread in the future is redirecting 'lbry://' links to a native LBRY application

That's something I may add in the future too.

ProgrammerWhoPrograms:
If there was only one public gateway that everyone used then IPFS would be centralised in practice.
I knew the comments were centralised but I didn't know search was. Thanks for letting me know. I've also heard rumours that most of the LBRY nodes are run by LBRY Inc but I don't know how true that is.
Still, that doesn't justify presenting Odysee as if it were LBRY.
It's a simple change that doesn't do much, so why not?

ProgrammerWhoPrograms:
IPFS gateways are centralised but IPFS itself isn't IMO.

FireMasterK:
> It's a simple change that doesn't do much, so why not?

- More people recognize LBRY over Odysee.
- I'm still using `api.lbry.tv` API, this can be seen at https://github.com/TeamPiped/Piped-Backend/blob/a15ad7081724e35c71887e6465f61db1504db69d/src/main/java/me/kavin/piped/utils/ResponseHelper.java#L925-L947

PR #414, opened 2021-09-05
Fix: Make more text translatable

ChunkyProgrammer:
This PR will make more of the text in the app translatable by adding new strings for text that doesn't already exist in the json file and referencing existing strings in places where the json file wasn't being used

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -18,7 +18,7 @@
             &nbsp;
             <a href="https://github.com/TeamPiped/Piped#donations">
                 <font-awesome-icon :icon="['fab', 'bitcoin']"></font-awesome-icon>
-                Donations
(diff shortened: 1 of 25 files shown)


FireMasterK:
Why did you rename `minimize_description` to `minimize_description_default`?

ChunkyProgrammer:
> Why did you rename `minimize_description` to `minimize_description_default`?

WatchVideo had text that was just "Minimize Description" and I feel like that suits the key better then "Minimize Description By Default". I can change it back if it's easier

FireMasterK:
I see, I just merged #415, please change it in Slovak/sk.json as well, and then I will merge this.

ChunkyProgrammer:
Done

FireMasterK:
Thanks!

PR #416, opened 2021-09-05
Add labels to improve accessibility

ChunkyProgrammer:
Adding labels to inputs/selects greatly increases accessibility by linking the text to the input element

Preference Page (before changes):
image

Preferences Page (after changes):
image

### DIFF
--- a/src/components/FeedPage.vue
+++ b/src/components/FeedPage.vue
@@ -18,8 +18,8 @@
     </span>
 
     <span class="uk-align-right">
-        <label for='ddlSortBy'>{{ $t("actions.sort_by") }}</label>
-        <select id='ddlSortBy' class="uk-select uk-width-auto" v-model="selectedSort" @change="onChange()">
+        <label for="ddlSortBy">{{ $t("actions.sort_by") }}</label>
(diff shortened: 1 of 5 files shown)


src/components/FeedPage.vue
```diff
@@ -18,8 +18,8 @@
     </span>
 
     <span class="uk-align-right">
-        {{ $t("actions.sort_by") }}
-        <select class="uk-select uk-width-auto" v-model="selectedSort" @change="onChange()">
+        <label for='ddlSortBy'>{{ $t("actions.sort_by") }}</label>
```
FireMasterK, line 21: You've used a mixture of single quotes and double-quotes. You should be using only double quotes. If you are using prettier, there is a config that you can import to do this when formatting :)

FireMasterK (changes requested):
Just had a look, everything else seems fine from a glance.

src/components/FeedPage.vue
ChunkyProgrammer: Sorry about that, should be fixed now

FireMasterK:
Thank you!

PR #417, opened 2021-09-05
fix: importing csv from youtube when not English

ChunkyProgrammer:
The headers in the csv ("Channel Id"/"Channel ID") can change depending on the user's language so I modified it to check to see if the file extension is 'csv' instead of whether the file contains "Channel Id"
Reference: https://github.com/TeamNewPipe/NewPipe/issues/6757#issuecomment-894696235


FireMasterK:
I see, interesting indeed! Thanks for the fix!

PR #418, opened 2021-09-06
Update languages with their native names

vachan-maker:
Usually for language selectors, all the languages are displayed in their native names.
So changed it from english to native name


FireMasterK:
Thanks!

PR #442, opened 2021-09-11
Add load more logic to feeds and fix feeds mobile design

IvRRimum:
I have a lot of subcsriptions, I had to wait like 30 seconds to load my feeds before, now happens instantly.


FireMasterK:
Instead of having a load more, I would rather have it infinitely scroll. Let me know if you would like me to do this, or if you're okay with doing it!

IvRRimum:
I will refactor in like an hour or so

IvRRimum:
Done

FireMasterK:
Thanks! I just made a few improvements for my satisfaction, so that nothing can break with this.

PR #464, opened 2021-09-20
translate countrymaps

ChunkyProgrammer:
This PR will make the list of countries translate to the user's selected language.

I created a script to change country.json files from https://github.com/stefangabos/world_countries to be formatted the same way as used in Piped. I don't know if this is the best way of implementing this feature so feel free to close this PR. If you would like me to share the format changing script then let me know. Thanks, ChunkyProgrammer


FireMasterK:
The only room for improvement I see is that they're in js files, they could be in JSON files technically from what I know.

ChunkyProgrammer:
> The only room for improvement I see is that they're in js files, they could be in JSON files technically from what I know.

Done

FireMasterK:
Perfect!

The other possibility I see is to use their npm package at https://www.npmjs.com/package/world_countries_lists so that we don't need to keep this list up to date, but that can be done later.

FireMasterK:
A regression from this - webpack now throws a lot of warnings when running `yarn serve`. The reason for this being the JSON is incorrect, with the last element terminating with a `,` which shouldn't be there.

ChunkyProgrammer:
oups, I am able to create a pr with a fix if you want

FireMasterK:
> oups, I am able to create a pr with a fix if you want

Sure thing!

PR #467, opened 2021-09-21
accessibility improvements

ChunkyProgrammer:
This PR makes Piped more accessible by bringing the # of accessibility errors down and also makes more elements translatable.

Trending Page (Before):
image

Trending Page (After):
image

Login Page (Before):
image

Login Page (After):
image

Preferences Page (Before):
image

Preferences Page (After):
image

### DIFF
--- a/src/components/VideoItem.vue
+++ b/src/components/VideoItem.vue
@@ -6,7 +6,7 @@
                 :width="width"
                 style="width: 100%"
                 v-bind:src="video.thumbnail"
-                alt="thumbnail"
+                alt=""
                 loading="lazy"
             />
             <div class="uk-position-relative">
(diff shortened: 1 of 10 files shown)


src/components/VideoItem.vue
```diff
@@ -30,14 +30,14 @@
         </router-link>
 
         <div class="uk-align-right" style="margin-left: 0; margin-bottom: 0; display: inline-block; width: 10%">
-            <router-link :to="video.url + '&listen=1'">
-                <font-awesome-icon icon="headphones"></font-awesome-icon>
+            <router-link :to="video.url + '&listen=1'" :aria-label="'listen to '+video.title">
+                <font-awesome-icon alt="" icon="headphones"></font-awesome-icon>
```
FireMasterK, line 34: I don't like the empty `alt`s, why exactly is this done? Couldn't something like `audio-only-button` or somethine else be used here for example?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt

here, it says:

> the alt text may be used by screen readers and other assistive technologies to help differently-abled users make full use of your content. It will be read aloud or sent to the touch device

ChunkyProgrammer: looks like font-awesome automatically adds "aria-hidden="true" " to icons so I don't think the empty alt actually does anything. The aria-label on the line before it tells screen readers what the icon does but i think it'd be better to add a title attribute as well.

Purpose of empty alts:
https://www.w3.org/WAI/tutorials/images/decorative/

The empty alt does make a lot more sense for the channel avatar on watchvideo, however

PR #474, opened 2021-09-23
Add support for better qualities through the IOS HLS manifest.

FireMasterK:
Implemented at https://github.com/FireMasterK/NewPipeExtractor/tree/ios-hls.

### DIFF
--- a/src/components/Player.vue
+++ b/src/components/Player.vue
@@ -195,6 +195,9 @@ export default {
                     method: "HEAD",
                 }).then(response => response.headers.get("Content-Type"));
                 mime = contentType;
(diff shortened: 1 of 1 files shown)


unixfox:
Do you want people that have iOS to test if this is working fine?

FireMasterK:
Sure, I just made the backend update! However, be aware that you will leak your IP to Google in case you try the link above.

unixfox:
From just viewing a couple of videos on both Safari for iOS and firefox for iOS everything is working fine.
I can't really test the video quality because I only have an iPhone SE 2020 (720p screen) on hand.

PR #494, opened 2021-10-03
Add hide recommendations button on mobile

### LINKED ISSUE #426: Button to hide recommended videos on mobile
Title.

DeeHechEss:
Closes #426 

I added a button similar to the hide description which is shown only on the single column grid (mobile) view to toggle the recommendations. I also added some logic to automatically display the recommendations if the screen increases in size.

Contribution for Hacktoberfest :)

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -122,7 +122,11 @@
             </div>
 
             <div class="uk-width-1-5@xl uk-width-1-4@s uk-width-1 uk-flex-last@s uk-flex-first" v-if="video">
+                <a class="uk-button uk-button-small uk-margin-small-bottom uk-hidden@s" style="background: #222" @click="showRecs = !showRecs">
+                    {{ showRecs ? $t("actions.minimize_recommendations") : $t("actions.show_recommendations") }}
+                </a>
                 <div
+                    v-show="showRecs || !smallView"
                     class="uk-tile-default uk-width-auto"
                     :style="[{ background: backgroundColor }]"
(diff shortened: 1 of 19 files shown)


src/components/WatchVideo.vue
```diff
@@ -144,6 +149,7 @@ import Comment from "@/components/Comment.vue";
 export default {
     name: "App",
     data() {
+        const smallViewQuery = window.matchMedia("(max-width: 640px)");
```
FireMasterK, line 152: Please move this to a computed variable.

FireMasterK (changes requested):

src/components/WatchVideo.vue
DeeHechEss: I tried as you suggested but I also had to initialize smallView to a boolean as it seems you can't access computed properties in data definition. Let me know if you had another way in mind.

FireMasterK: I had thought about removing `smallViewQuery` and replacing it with `smallView` as a computed variable!

DeeHechEss: I tried that but it wouldn't update itself when I resized the window. I could use forceUpdate in the event listener instead though.

`    computed: {
        smallView() {
            return window.matchMedia("(max-width: 640px)").matches;
        }
    },`

FireMasterK: Oh okay, then your latest commit was unnecessary then, I'll revert it and merge this!

FireMasterK (approved):

PR #495, opened 2021-10-04
Fix safari fullscreen

### LINKED ISSUE #394: Full screen doesn't full screen on Safari 14 macOS
Full screen doesn't full screen on Safari 14 macOS.

Safari
safari

Firefox
firefox

Chromium
chromium

DeeHechEss:
Closes #394

The issue was the `max-height: 75vh` set when `isEmbed` is true is not overridden by Safari in fullscreen automatically.

Contribution for Hacktoberfest :)


FireMasterK:
I see, thanks!

PR #508, opened 2021-10-07
feat: adds the option to toggle listen mode when watching a video

filipesmedeiros:
i don't know if this is the most efficient way of doing this lol

first time coding in vue, let me know if anything can be improved

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -55,8 +55,8 @@
                 </a>
                 <router-link
                     :to="toggleListenUrl"
-                    :aria-label="(isListening ? 'watch ' : 'listen to ') + video.title"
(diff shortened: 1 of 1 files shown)


filipesmedeiros:
i'd also love to improve one thing: when clicking the link, staying on the same time

i've checked and you support the youtube convention of `?t=...` which is awesome. that would mean that in some way we could get the current time when clicking the link and using that for the query of the "new url". only issue i see with this is imperative routing instead of declarative, but i don't think it's a big issue?

FireMasterK:
> i've checked and you support the youtube convention of ?t=... which is awesome. that would mean that in some way we could get the current time when clicking the link and using that for the query of the "new url". only issue i see with this is imperative routing instead of declarative, but i don't think it's a big issue?

Create a new issue for it :)

PR #518, opened 2021-10-10
feat: menu and feed start

filipesmedeiros:
i think it's pretty neat, but you're welcome to test it and give some suggestions!

i tested it and it works pretty well on desktop sizes and mobile

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,7 +7,7 @@
         />
         <main
             class="uk-container uk-container-expand"
-            style="height: 100vh; overflow: scroll;"
+            style="height: 100vh; overflow: scroll; flex: 1;"
             :style="{ background: backgroundColor, colour: foregroundColor, marginTop: isMobile ? '70px' : 0 }"
             :class="{ 'uk-light': darkMode }"
         >


FireMasterK:
The only problem I see is that the build fails due to `sass-loader` not being present! :thinking:

filipesmedeiros:
this is weird, it was a mistake on my part that is unintended! i will fix it asap! 😄

filipesmedeiros:
@FireMasterK you can check it! seems nice now 😄

FireMasterK:
For some odd reason, some menus have a weird black bar and don't occupy the entire screen :thinking: 

image

filipesmedeiros:
yeah! this pr only handles the menu (and the trending page because it was easier

i will pr that asap, but wanted to keep them fairly separated.

i can fix this one thing though, seems easy enough

FireMasterK:
Okay, it happened in the feed and trending page too, which is why I thought it was wierd! (before it loads)

filipesmedeiros:
done!

i'd like to add a loading state eventually

PR #530, opened 2021-10-12
feat: no animations

filipesmedeiros:
@mhowser


Mhowser:
Can you make this a toggle in settings?

filipesmedeiros:
no one's ever gonna enable them though ahah (i don't mind just turning them off completely tbh)

Mhowser:
Yeah I agree, better to have them on by default because most modern apps have nice animations. Its good to allow disabling them for low end devices.

Issue #618, opened 2021-11-12
Add Shortcut Key "C" for Captions in the video to toggle it on or off.


Loki-L1130:

FireMasterK:
Closing this with f1f45551f42ab0101d75b648a9706a9e0a784420.

PR #627, opened 2021-11-15
Add hotkey to enable captions with "c"

metzgegu:
Enable/Disable captions visibility when pressing the `c` key

Related Issue : #618


FireMasterK:
Works perfectly, thank you!

PR #628, opened 2021-11-16
support path in proxyURL

aspacca:


src/components/Player.vue
```diff
@@ -186,8 +186,11 @@ export default {
                 uri = lbry.url;
                 if (this.getPreferenceBoolean("proxyLBRY", false)) {
                     const url = new URL(uri);
+                    const proxyURL = new URL(this.video.proxyUrl);
                     url.searchParams.set("host", url.host);
-                    url.host = new URL(this.video.proxyUrl).host;
+                    url.port = proxyURL.port;
```
FireMasterK, line 191: This is redundant:
```
new URL("https://www.youtube.com:8443/test").host
```
returns `'www.youtube.com:8443'`

src/components/Player.vue
```diff
@@ -219,10 +222,12 @@ export default {
                                 (component.getPreferenceBoolean("proxyLBRY", false) || headers.Range))
                         ) {
                             url.searchParams.set("host", url.host);
-                            url.host = proxyHost;
+                            url.port = proxyURL.port;
```
FireMasterK, line 225: Same here

src/components/Player.vue
```diff
@@ -186,8 +186,11 @@ export default {
                 uri = lbry.url;
                 if (this.getPreferenceBoolean("proxyLBRY", false)) {
                     const url = new URL(uri);
+                    const proxyURL = new URL(this.video.proxyUrl);
                     url.searchParams.set("host", url.host);
-                    url.host = new URL(this.video.proxyUrl).host;
+                    url.port = proxyURL.port;
+                    url.host = proxyURL.host;
+                    url.pathname = proxyURL.pathname + url.pathname;
```
FireMasterK, line 193: This causes problems as there would be a double slash since:
```
new URL("https://www.youtube.com:8443").pathname
```
returns `'/'`

src/components/Player.vue
```diff
@@ -219,10 +222,12 @@ export default {
...
                         ) {
                             url.searchParams.set("host", url.host);
-                            url.host = proxyHost;
+                            url.port = proxyURL.port;
+                            url.host = proxyURL.host;
+                            url.pathname = proxyURL.pathname + url.pathname;
                             request.uris[0] = url.toString();
                         }
-                        if (url.pathname === "/videoplayback") {
+                        if (url.pathname === proxyURL.pathname + "/videoplayback") {
```
FireMasterK, line 230: Same problem as above.

FireMasterK (changes requested):

src/components/Player.vue
```diff
@@ -214,6 +212,7 @@ export default {
 
                     const localPlayer = new this.shaka.Player(videoEl);
                     const proxyURL = new URL(component.video.proxyUrl);
+                    const proxyPath = proxyURL.pathname === "/" ? "" : proxyURL.pathname;
```
FireMasterK, line 215: I think you intend this to be `proxyURL.slice(0, -1)`

aspacca: that will return `""` in case of `/` but `path/with/leading/slash` in case of `/path/with/leading/slash`, what would be wrong

FireMasterK: But then you add it to URL.pathname, which starts with a /?

You'd end up with a // then

aspacca: but `/` case is covered and `proxyPath` will be an empty string

`/path/with/leading/slash + /url/pathname`
or ` + /url/pathname`

your solution will end with
`path/with/leading/slash + /url/pathname`

FireMasterK: But let's say `proxyUrl = new URL("https://piped.kavin.rocks/proxy/")`

You'll have problems, here, right?

aspacca: we want to have `uri.pathname = "proxy" + "/url/pathname"`, oki
but what if I have `proxyUrl = new URL("https://piped.kavin.rocks/proxy")`?

can we assume that an `uri.pathname` with no leading slash will build the correct url?

aspacca: will fix ASAP

aspacca: done

src/components/Player.vue
FireMasterK: I think this just has to be `proxyPath`, and then it looks good.

aspacca: fixed it

aspacca:
@FireMasterK anything else to change?

PR #646, opened 2021-11-24
Added Serbian translations from Weblate

milutinke:
Translated the project into Serbian language on Web Late and included the translation into the project and added 'CountryMaps' for Serbian.


FireMasterK:
Thank you!

PR #660, opened 2021-12-04
Play playlist in order

holger-frovin:
When watching video from playlist, by either clicking on video from playlist component or by using link with link parameter, plays videos in order and shows rest of playlist as related videos in order.

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -131,7 +131,7 @@
                     {{ showRecs ? $t("actions.minimize_recommendations") : $t("actions.show_recommendations") }}
                 </a>
                 <div
-                    v-for="related in video.relatedStreams"
+                    v-for="related in isPlaylist ? playlist.relatedStreams : video.relatedStreams"
                     v-show="showRecs || !smallView"
                     :key="related.url"
                     class="uk-tile-default uk-width-auto"
@@ -165,6 +165,9 @@ export default {
             video: {
                 title: "Loading...",
             },
+            playlist: {
+                title: "Loading...",
+            },
             sponsors: null,
             selectedAutoLoop: false,
             selectedAutoPlay: null,
@@ -189,6 +192,9 @@ export default {
         },
         isEmbed(_this) {
             return String(_this.$route.path).indexOf("/embed/") == 0;
+        },
+        isPlaylist(_this) {
+            return _this.$route.query.list;
         },
         uploadDate(_this) {
             return new Date(_this.video.uploadDate).toLocaleString(undefined, {
@@ -228,6 +234,7 @@ export default {
             })();
             if (this.active) this.$refs.videoPlayer.loadVideo();
         });
+        this.getPlaylistData();
         this.getSponsors();
         if (!this.isEmbed && this.getPreferenceBoolean("comments", true)) this.getComments();
         window.addEventListener("resize", () => {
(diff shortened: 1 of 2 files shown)


FireMasterK:
Hey, thanks for the pull request but there are a few un-addressed problems I see in your approach:
- When "Auto Play next Video" is enabled, it plays a recommended video
- An index query could be added to tell what video you are currently on
- Continuations for playlist are not handled (both in the Playlist and WatchVideo files)

holger-frovin:
I'm pretty sure I understood what the first two items mean and I think i have implemented them now, but I'm not sure what you mean by Continuations.

SamantazFox:
> I'm pretty sure I understood what the first two items mean and I think i have implemented them now, but I'm not sure what you mean by Continuations.

"Continuations" (also know as ctokens) are base64-encoded protobuf objects that have to be passed along to the innertube API to get the next page of content (being for a playlist, search results, channel videos, etc...). Without those, the results returned will always be the first 100 videos of a playlist, the 50 last videos of a channel, etc...

holger-frovin:
Oops I just found an error in autoplay. Will fix.

src/components/VideoItem.vue
```diff
@@ -96,6 +103,14 @@ export default {
                 return {};
             },
         },
+        playlist: {
+            type: Object,
+            default: () => {
+                return {};
+            },
+        },
+        isPlaylist: Boolean,
```
FireMasterK, line 112: These changes may be unnecessary since they're not required if the above comments are addressed.

src/components/WatchVideo.vue
```diff
@@ -329,6 +357,34 @@ export default {
             this.subscribed = !this.subscribed;
         },
         handleScroll() {
+            if (this.loading || !this.isPlaylist || !this.playlist.nextpage) return;
+            if (window.innerHeight + window.scrollY >= document.body.offsetHeight - window.innerHeight) {
```
FireMasterK, line 361: This would potentially require the user to scroll to the end of the page, instead of just till the end of the playlist.

src/components/WatchVideo.vue
```diff
@@ -329,6 +357,34 @@ export default {
             this.subscribed = !this.subscribed;
         },
         handleScroll() {
+            if (this.loading || !this.isPlaylist || !this.playlist.nextpage) return;
```
FireMasterK, line 360: Won't this break the loading of more comments if we're not in a playlist?

src/components/Player.vue
```diff
@@ -328,13 +335,30 @@ export default {
                 videoEl.addEventListener("ended", () => {
                     if (!this.selectedAutoLoop && this.selectedAutoPlay && this.video.relatedStreams.length > 0) {
                         const params = this.$route.query;
-                        let url = this.video.relatedStreams[0].url;
+                        let index = Number(params.index).valueOf();
                         const searchParams = new URLSearchParams();
+                        let playlistEnded = index == this.playlist.videos;
+                        let url =
+                            this.isPlaylist && playlistEnded
```
FireMasterK, line 342: I think you meant this to be `!playlistEnded` :confused:

FireMasterK (changes requested):
Some minor improvements could be made, also when autoplay of the next video is enabled, it plays a recommended video.

PR #694, opened 2021-12-23
Add support for displaying chapters

### LINKED ISSUE #604: Chapter Support
This is something that I really miss from YouTube!

FireMasterK:
Closes #604


opusforlife2:
Chapter support works great! 🎉  I was wondering if you could sync up the behaviour of description timestamps with chapters. Right now, clicking on timestamps reloads the whole page and starts the stream from that timestamp. But if it worked like clicking on chapters does, simply seeking to the correct time, that would be awesome.

PR #683, opened 2021-12-18
Replace UIKit with Windicss

FireMasterK:
Closes #706
Closes #703

PR #701, opened 2021-12-28
Refactor template code to increase readability

kreha1:
This is a minor refactor done in order to ease the transition to windicss. I wish to rewrite the whole project with typescript, fully transition to the composition api using script setup and convert mixins to composable functions.

I'd also like to move from webpack to [vite](https://vitejs.dev/guide/why.html#slow-server-start), as this is where all the nice things are at this moment (like pretty much instant dev server start, and plenty of plugins), as well as to move from yarn (slower, FB owned) to [pnpm](https://pnpm.io/).

I wanted to limit changes to mostly HTML, but there are many places where logic itself could be improved by using optional chaining, nullish coalescing operator, or simply using await keyword instead of assigning values inside `then` callback.

If you are okay with my contribution, let me know in the comment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
         "serve": "vite",
         "build": "vite build",
         "preview": "vite preview",
-        "lint": "eslint --fix --color --ignore-path .gitignore ."
+        "lint": "eslint --fix --color --ignore-path .gitignore --ext .js,.vue ."
     },
     "dependencies": {
         "@fortawesome/fontawesome-svg-core": "^1.2.36",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
     "scripts": {
         "serve": "vite",
         "build": "vite build",
-        "preview": "vite preview"
+        "preview": "vite preview",
+        "lint": "eslint --fix --color --ignore-path .gitignore ."
     },
     "dependencies": {
         "@fortawesome/fontawesome-svg-core": "^1.2.36",
@@ -31,10 +32,12 @@
         "@vue/compiler-sfc": "3.2.26",
         "babel-eslint": "^10.1.0",
         "eslint": "^7.32.0",
+        "eslint-config-prettier": "^8.3.0",
         "eslint-plugin-prettier": "^4.0.0",
         "eslint-plugin-vue": "^7.20.0",
         "prettier": "^2.5.1",
         "vite": "^2.7.9",
+        "vite-plugin-eslint": "^1.3.0",
         "vite-plugin-pwa": "^0.11.12",
         "vite-plugin-windicss": "^1.6.1"
     },
--- a/src/components/Playlist.vue
+++ b/src/components/Playlist.vue
@@ -8,16 +8,16 @@
 
         <div class="grid grid-cols-2">
             <div>
-                <b
+                <strong
                     ><router-link class="uk-text-justify" :to="playlist.uploaderUrl || '/'">
                         <img :src="playlist.uploaderAvatar" loading="lazy" class="rounded-full" />
                         {{ playlist.uploader }}</router-link
-                    ></b
+                    ></strong
                 >
             </div>
             <div>
                 <div class="right-2vw absolute">
-                    <b>{{ playlist.videos }} {{ $t("video.videos") }}</b>
+                    <strong>{{ playlist.videos }} {{ $t("video.videos") }}</strong>
                     <br />
                     <a :href="getRssUrl"><font-awesome-icon icon="rss"></font-awesome-icon></a>
                 </div>
--- a/src/components/Playlist.vue
+++ b/src/components/Playlist.vue
@@ -6,12 +6,10 @@
 
         <div class="grid grid-cols-2">
             <div>
-                <strong>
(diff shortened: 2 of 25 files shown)


src/components/Playlist.vue
```diff
@@ -2,24 +2,24 @@
...
             <div>
-                <b
-                    ><router-link class="uk-text-justify" :to="playlist.uploaderUrl || '/'">
+                <strong>
+                    <router-link class="uk-text-justify" :to="playlist.uploaderUrl || '/'">
                         <img :src="playlist.uploaderAvatar" loading="lazy" class="rounded-full" />
-                        {{ playlist.uploader }}</router-link
-                    ></b
-                >
+                        <span v-text="playlist.uploader" />
```
kreha1, line 12:
```suggestion
                        <strong v-text="playlist.uploader" />
```

src/components/Playlist.vue
```diff
@@ -2,24 +2,24 @@
...
-        <h1 class="text-center">
-            {{ playlist.name }}
-        </h1>
+        <h1 class="text-center" v-text="playlist.name" />
 
         <div class="grid grid-cols-2">
             <div>
-                <b
-                    ><router-link class="uk-text-justify" :to="playlist.uploaderUrl || '/'">
+                <strong>
```
kreha1, line 9:
```suggestion
```

src/components/Playlist.vue
```diff
@@ -2,24 +2,24 @@
...
+                    <router-link class="uk-text-justify" :to="playlist.uploaderUrl || '/'">
                         <img :src="playlist.uploaderAvatar" loading="lazy" class="rounded-full" />
-                        {{ playlist.uploader }}</router-link
-                    ></b
-                >
+                        <span v-text="playlist.uploader" />
+                    </router-link>
+                </strong>
```
kreha1, line 14:
```suggestion
```

kreha1 (commented):
I've tracked some leftovers.

FireMasterK:
Thank you for the pull request!

> I wish to rewrite the whole project with typescript, fully transition to the composition api using script setup and convert mixins to composable functions.

I'm not familiar with typescript, but I'm willing to learn :)

> I'd also like to move from webpack to vite

This would be nice

> as well as to move from yarn to pnpm

What happens to dependabot then? I don't think it supports pnpm sadly

> optional chaining, nullish coalescing operator

I believe these require webpack 5 :thinking:, not sure what happens if we use vite tho

> If you are okay with my contribution, let me know in the comment.

Yes, I had a look at these changes, it looks good to me!

package.json
```diff
@@ -29,15 +29,16 @@
...
+        "@vue/eslint-config-prettier": "^6.0.0",
         "babel-eslint": "^10.1.0",
-        "eslint": "^7.32.0",
-        "eslint-config-prettier": "^8.3.0",
-        "eslint-plugin-prettier": "^4.0.0",
-        "eslint-plugin-vue": "^7.20.0",
+        "eslint": "^6.7.2",
+        "eslint-plugin-prettier": "^3.3.1",
+        "eslint-plugin-vue": "^7.0.0",
+        "prettier": "^2.2.1",
```
FireMasterK, line 41: Why are we downgrading some dependencies?
eg: eslint `7.32.0` -> `6.7.2`

kreha1: npm and pnpm couldn't resolve dependencies when I first tried fetching packages, due to `@vue/cli-plugin-eslint@4.5.15` having set `eslint@">= 1.6.0 < 7.0.0"` as a peer dependency. Yarn doesn't actually care, so I can as well revert this commit. Especially since when will have switched to vite, this dependency will no longer be there.

Would you prefer a commit reverting the changes, or to leave it out during a rebase? The latter will require a force push, but that won't have any implications beyond this PR.

FireMasterK: A revert works just as fine :)

kreha1:
> > as well as to move from yarn to pnpm
>
> What happens to dependabot then? I don't think it supports pnpm sadly

So this will be on-hold until they do support it. This should be a matter of time, as [Renovate](https://github.com/renovatebot/renovate/) already supports pnpm.

> > optional chaining, nullish coalescing operator
> 
> I believe these require webpack 5 🤔, not sure what happens if we use vite tho

Vite uses ES modules by default, so it's only a matter of configuring babel if it's even needed as [Optional Chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) and [Nullish Coalescing Operator](https://caniuse.com/mdn-javascript_operators_nullish_coalescing) have been supported by all major browser for more than a year already.

FireMasterK:
I've just migrated from Webpack to Vite in https://github.com/TeamPiped/Piped/pull/683/commits/562f139ad548c5375846fd8afeac2934ba0b8ca2, however, some work is needed to get PWA support and eslint working.

kreha1:
I have rebased my branch onto your vite commit, so there are no conflicts (I've also dropped the first commit which changed eslint version). I've additionally made it so it passes the github action check, as when replacing vue-cli with vite, `lint` script in package.json had to be added manually.

If there are no issues on the part of my code, please merge it, so I can focus on other improvements ;).

PR #726, opened 2022-01-13
Wait to save video timestamp until saved video timestamp is loaded

ishehadeh:
Fixes a race condition where updateProgressDatabase() would run before the saved time could be loaded.
Since the initial timestamp on the video element was 0 the saved time would be forgotten.


FireMasterK:
Thank you!

PR #863, opened 2022-02-26
Make URLs in video description links

Jihakuz:
Currently, if a video has links in it's description you are unable to click and follow the link.

For example:
[Piped](https://piped.kavin.rocks/watch?v=Qt_-nVOGjKo)
[Invidious](https://inv.riverside.rocks/watch?v=Qt_-nVOGjKo)
[YouTube](https://www.youtube.com/watch?v=Qt_-nVOGjKo)

To do this I just parse the video's description through a `urlify` function I found laying around. Thinking about this now; is it possible to run this parsing when this `video` object (? I couldn't find where it actually came from, still new to Vue) is created?

This might eventually want to be expanded to hashtags as well.

This is my first time messing about with Vue/contributing to this project, so sorry if I've done something wrong/against common conventions.

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -89,7 +89,7 @@
                 v-t="`actions.${showDesc ? 'minimize_description' : 'show_description'}`"
             />
             <!-- eslint-disable-next-line vue/no-v-html -->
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thanks for the contribution! I've just moved the function away from rendering and prevented it from re-rewriting descriptions that already have HTML links.

PR #899, opened 2022-03-14
Move all controls in overflow menu to main panel

nityy:
These changes might be subjective but now it takes one less step to change controls. Enabled tooltips so current applied values can be seen with hover, except for mobile because tooltips are hard to trigger on mobile and might actually be annoying.

Screenshots:

Screenshot from 2022-03-14 21-26-35

Screenshot from 2022-03-14 21-18-25

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -416,18 +416,30 @@ export default {
                     }
                 };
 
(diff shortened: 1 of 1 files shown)


Bnyro:
Closing because it should be optional only since it's a mostly matter of personal taste and the current way of handling it doesn't have any deal-breaking issues.

PR #915, opened 2022-03-17
fix the broken channel image height

luphoria:
super simple fix, the image height is initially set to `auto` but that causes this issue when viewing subscription at the wrong aspect ratio:
image
This fix simply changes the image height adjust for channels to `fit-content`, which fixes the icons to look more like this:
image
:bread: Much love :smiling_face_with_three_hearts:

### DIFF
--- a/src/components/SubscriptionsPage.vue
+++ b/src/components/SubscriptionsPage.vue
@@ -15,7 +15,7 @@
             <div class="flex justify-center place-items-center">
                 <div class="w-full grid grid-cols-3">
                     <router-link :to="subscription.url" class="col-start-2 block flex text-center font-bold text-4xl">
-                        <img :src="subscription.avatar" class="rounded-full" width="48" height="48" />
+                        <img :src="subscription.avatar" class="rounded-full channel" width="48" height="48" />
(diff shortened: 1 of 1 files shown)


src/components/SubscriptionsPage.vue
```diff
@@ -15,7 +15,7 @@
             <div class="flex justify-center place-items-center">
                 <div class="w-full grid grid-cols-3">
                     <router-link :to="subscription.url" class="col-start-2 block flex text-center font-bold text-4xl">
-                        <img :src="subscription.avatar" class="rounded-full" width="48" height="48" />
+                        <img :src="subscription.avatar" class="rounded-full channel" width="48" height="48" />
```
FireMasterK, line 18: The channel class could just be replaced to `h-[fit-content]`

Alternatively, a class for channel could be created:
```
.channel {
    @apply rounded-full h-[fit-content];
}
```

These are just WindiCSS directives :)

luphoria: Nice :P I am a total vue / windicss noob so I tried to stick with base HTML+CSS. I honestly didn't think of just changing height from 48 there lol

PR #1060, opened 2022-05-16
show message if comments are disabled in a video

laazyCmd:
i think it would be nice to show something if comments are disabled in a particular video for QOL, i also added an indicator if comments are still loading.

**Preview**
if-comments-disabled


FireMasterK:
Thank you!

PR #1136, opened 2022-06-16
Improve Margins and add Theater Mode

maltejur:
Improve a few margins. I know there is a UI rework coming but this should be just a small quick change. Most notably, add a margin to the left and right of the page if a video is playing.

Before:
Screenshot 2022-06-16 at 21-31-52 Piped

After (note that the video has pretty much the same space due to the black bars):
Screenshot 2022-06-16 at 21-32-04 Piped

Before:
Screenshot 2022-06-16 at 21-32-41 Piped

After:
Screenshot 2022-06-16 at 21-32-29 Piped

Let me know what you think. If you want to merge the page margin at the left and right, I could also implement a "theater mode" that reverts it again. If not, I could also exclude the page margin from this MR.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -63,6 +63,21 @@ export default {
             }
         })();
     },
+    watch: {
+        $route: {
+            handler(to) {
+                console.log(to.name);
+                // Pages that should be displayed without margin at the sides
+                if (
+                    ["Trending", "SearchResults", "Playlist", "Channel", "Feed", "Watch History", "Playlists"].includes(
+                        to.name,
+                    )
+                )
+                    document.body.classList.add("wide");
+                else document.body.classList.remove("wide");
+            },
+        },
+    },
 };
 </script>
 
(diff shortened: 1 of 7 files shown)


FireMasterK:
> I could also implement a "theater mode" that reverts it again.

I would love that since I love how currently the full screen's real estate is used.

maltejur:
Okay, done. Also got a quick preview here: https://piped-80dr7mank-maltejur.vercel.app/

matissime (approved):
Great feature that work well, hope to see it coming!

kskarthik:
@FireMasterK what's the blocker here?

FireMasterK:
Some problems/things I noticed in my testing:
- When loading the home/other pages, it would initially flicker from the half-screen layout to full screen
- The logo, feed/login buttons on the top are in the center
- I don't like this being the default personally, but I could create a poll to see what the community wants.

PR #1175, opened 2022-07-02
feat: add rss feed button for channels

kskarthik:
## UI 

image


FireMasterK:
I'm not a fan of linking directly to YouTube, but this is okay as a temporary solution.

kskarthik:
> I'm not a fan of linking directly to YouTube, but this is okay as a temporary solution.

@FireMasterK newpipe also does the same ATM, Let's see if we can find a better way

Issue #1194, opened 2022-07-10
Toggle Audio/Video button on playback page forwards to wrong URL


honey-roast-coconut:
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.kavin.rocks/) or is API related.

### Describe the bug

https://user-images.githubusercontent.com/109007873/178129428-12673e73-150f-4cd0-b8f7-7fc0d7f043ca.mp4

![image](https://i.gyazo.com/38a386c74326bac6a740befe20182555.png)

### To Reproduce

When clicking headphone/display button

### Expected behavior

Just takes to audio/video version page

### Logs/Errors

No errors found on dev console.

### Browser, and OS with Version.

Windows Edge 103.0.1264.49
Android Chrome 103.0.5060.71

kskarthik:
I observed this too

PR #1200, opened 2022-07-11
Use a relative url with router-link (fixes: #1194)

Ikkibird:
Good evening,

First of all, thanks for your work on this amazing project!

I recently stumbled across a bug while using the "listen" feature on a video (#1194). After a few researches, it seems that the `toggleListenUrl` function return an absolute url path to the `<router-link>` component, which can only accept relative paths (vuejs/vue-router#1131).

So I fixed it by making it return `url.pathname + url.search` instead, which is the concatenation of the relative path and the URL query parameters.

As this is my first pull request ever, feel free to correct me if I have done anything wrong. :wink:

Regards,

Ikkibird

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -251,7 +251,7 @@ export default {
         toggleListenUrl(_this) {
             const url = new URL(window.location.href);
             url.searchParams.set("listen", _this.isListening ? "0" : "1");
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thank you, this is perfect!

kskarthik:
Wow!

Issue #1140, opened 2022-06-17
[Feature Request] option to disable recommended/related videos feed


SiddharthSharma16923:
reduce , the no. of requests handled by piped proxy and the user is happy
win-win

one more
> default speed of video option

FireMasterK:
Closing this with #1218.

Issue #806, opened 2022-02-08
Hide Videos On Homepage, Recommended Videos Settings


trymeouteh:
To make Youtube less addictive, please add settings to show/hide the following...

- Videos on Homepage
- Reccomended Videos (Videos along the side of a video page)

Newpipe app has these settings and so does Invidious. These settings can help with making the platform less addictive.

Loki-L1130:
For the first one. I think Subscribing to some channels and setting the Homepage as feeds will help you.

For recommended videos alongside, video. For that, @FireMasterK made it alongside the comments.

trymeouteh:
The "Show Comments" setting only enable/disables the comments. It would be nice to have a settings to hide the videos that are alongside where the comments appear.

FireMasterK:
Closing this with #1218, you can't hide the homepage, however, but you can change it by default to show your feed.

Issue #521, opened 2021-10-11
A couple feature requests


travilingyoshi:
I like piped, especially compared to invidious, but I have a few feature requests that I think could be good for the service.
- [ ] Theater mode: one feature that's missed from both youtube and invidous is the toggle-able theater mode that makes the screen size wider. This is the feature I want most out of the layout, if possible.
- [ ] reddit comments: toggle-able reddit comments could be nice, but not necessary if not possible to implement.
- [ ] right click sharing. Youtube has a feature where you can right click the video and get some options to share (screnshot example: 
![image](https://user-images.githubusercontent.com/92283253/136725617-f1bbf2d1-fc3d-4860-b78a-a7294e81d86b.png)
). If not possible, maybe a share button somewhere below the title that serves the same function.
- [x] Work mode: an option to hide front page and side recommended videos, similar to the extension DF youtube (https://chrome.google.com/webstore/detail/df-tube-distraction-free/mjdepdfccjgcndkmemponafgioodelna)

Bnyro:
Please open one issue per feature request, closing this one as it's hard to track.

PR #1218, opened 2022-07-19
Option to minimize recommendations on desktop and by default

macks36:
This makes the "Show Recommendations" button visible on desktop and adds a preference to minimize them by default.
This has been mentioned in #521, #806 and #1140. While these issues suggest to completely hide the recommendations, I think having an option to minimize them by default makes more sense here since there already is a button to hide them, it’s just hidden on desktop.

Because there is new text for the preference now, a new key for the translation is needed. I suppose this is added through weblate?

### DIFF
--- a/src/components/PreferencesPage.vue
+++ b/src/components/PreferencesPage.vue
@@ -124,6 +124,16 @@
         @change="onChange($event)"
     />
     <br />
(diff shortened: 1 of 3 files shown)


FireMasterK:
Thank you! I have added the missing translation string.

PR #1227, opened 2022-07-20
Replaces Privacy Redirect with Libredirect

Dyras:
Privacy Redirect hasn't been actively updated in almost a year now (25 Jul, 2021) so I think it should be replaced with Libredirect which is very much updated almost weekly.

There's a Firefox version, and an Edge version. The Edge version is unfortunately taken down (falsely) by Microsoft due to "malware": https://github.com/libredirect/libredirect/issues/269


FireMasterK:
Thank you!

PR #1230, opened 2022-07-20
make trending accessible although feed is the default tab

Bnyro:
I don't know whether that's a good idea but it would allow users to browse the trending tab although they selected the feed as default tab. Currently there is no way to watch the trending videos when the default tab is set to feed.

### DIFF
--- a/src/components/TrendingPage.vue
+++ b/src/components/TrendingPage.vue
@@ -31,14 +31,17 @@ export default {
     activated() {
         document.title = this.$t("titles.trending") + " - Piped";
         if (this.videos.length > 0) this.updateWatched(this.videos);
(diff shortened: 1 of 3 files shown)


src/components/TrendingPage.vue
```diff
@@ -31,14 +31,17 @@ export default {
...
         if (this.videos.length > 0) this.updateWatched(this.videos);
-        switch (this.getPreferenceString("homepage", "trending")) {
-            case "trending":
-                break;
-            case "feed":
-                this.$router.push("/feed");
-                return;
-            default:
-                break;
+        console.log(this.$route.path);
```
FireMasterK, line 34: Remove this

Bnyro: whoops :p

PR #1231, opened 2022-07-20
log out all devices

### LINKED ISSUE #1183: Invalidating sessions and deleting accounts
Provide a way to logout all other sessions along with ability to delete account like Libretube does in v0.3.3

Bnyro:
closes #1183

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -82,9 +82,9 @@
         "delete_playlist_confirm": "Are you sure you want to delete this playlist?",
         "please_select_playlist": "Please select a playlist",
         "delete_account": "Delete Account",
-        "logout": "Log out this device",
+        "logout": "Logout from this device",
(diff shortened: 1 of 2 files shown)


src/locales/en.json
```diff
@@ -82,8 +82,9 @@
         "delete_playlist_confirm": "Are you sure you want to delete this playlist?",
         "please_select_playlist": "Please select a playlist",
         "delete_account": "Delete Account",
-        "logout": "Logout",
-        "minimize_recommendations_default": "Minimize Recommendations by default"
+        "logout": "Log out this device",
+        "minimize_recommendations_default": "Minimize Recommendations by default",
+        "invalidate_session": "Log out all devices"
```
FireMasterK, line 87: I would have "Logout" as one word and, and add "from" after it. Just a nit-pick :p

PR #1236, opened 2022-07-21
clone playlist from youtube

Bnyro:
I couldn't find a better way to figure out whether a playlist is a Piped playlist or not, but I'm not sure whether YouTube playlists can contain "-" too.


FireMasterK:
You can just test if they match this regex: `[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`

PR #1340, opened 2022-08-27
Block channels

### LINKED ISSUE #1090: [Request] Block Channels
It would be nice to be able to block channels from your view. I would envision another icon (possibly an eye) next to audio only, and add to playlist that would add that channel name to a block list which is auto filtered when searching, viewing the trending section, or any related videos while watching a video. This list could be managed under the preferences for the account.

image

Bnyro:
closes #1090


montchr:
It looks like this was abandoned? If so, why? 

https://github.com/libre-tube/LibreTube/issues/1173 mentions the existence of this pull request but it seems that is no longer relevant when the trail stops here.

FireMasterK:
> It looks like this was abandoned? If so, why?

There were issues in the implementation (eg: using localStoage instead of IndexedDb, which can cause performance issues when large number of channels are blocked), and this wasn't really a priority to add support for.

PR #1343, opened 2022-08-28
Share Dialog

### LINKED ISSUE #886: Generate URL with custom timestamp
Please add a button which will open a modal popup which you can use the current timestamp of the video/audio or set the timestamp to a custom time. It will then provide a youtube.com URL and a URL for the current Piped domain. And by clicking on the generated link will reload the video/audio at the timestamp for quick and easy bookmarking.

Bnyro:
replaces the "watch on YT" button
closes #886

### DIFF
--- a/src/components/ModalComponent.vue
+++ b/src/components/ModalComponent.vue
@@ -1,0 +1,42 @@
+<template>
+    <div class="modal">
+        <div>
+            <div class="modal-container">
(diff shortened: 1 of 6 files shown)


FireMasterK:
What we should do is refractor this to add a model, we're adding unnecessary duplicate code!

PR #1371, opened 2022-09-02
fix timestamp in share links

felschr:
Piped links created via the share dialogue contain 2 query parameters, but both are prefixed with `?` instead of using `&` for subsequent parameters.
This wasn't an issue for YouTube links, since they don't use any other query parameters other than the timestamp.

To handle both cases, I used [`URL.searchParams.append()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/append) in my solution.

### DIFF
--- a/src/components/ShareModal.vue
+++ b/src/components/ShareModal.vue
@@ -69,11 +69,12 @@ export default {
     },
     computed: {
         generatedLink() {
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thank you!

PR #1537, opened 2022-10-06
Better handling of the auto theme to prevent further bugs with it

Bnyro:


src/main.js
```diff
@@ -239,6 +239,13 @@ const mixin = {
     },
     computed: {
         theme() {
+            this.refreshTheme; // forces Vue to recompute the value when the var gets changed
+            if (this.getPreferenceString("theme", "dark") == "auto") {
+                const darkModePreference = window.matchMedia("(prefers-color-scheme: dark)");
+                // Forces the app to recompute the theme class
+                darkModePreference.addEventListener("change", () => this.refreshTheme++);
+                return darkModePreference.matches ? "dark" : "light";
+            }
```
FireMasterK, line 248: When `theme()` is computed multiple times, there will be multiple event listeners registered.

PR #1550, opened 2022-10-09
fix: align buttons

### LINKED ISSUE #1549: Buttons have different height
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.kavin.rocks/) or is API related.

### Describe the bug

In the official hosted instance, each button has a different size making them all inconsistent

### To Reproduce

1. Click on a video
2. See Subscribe, rss, share, and audio/video toggle - each has a different height

### Expected behavior

All buttons should have a consistent height

### Logs/Errors

-

_No response_

### Additional context

Screen Shot 2022-10-09 at 8 56 03

talkor:
Resolves #1549

Screen Shot 2022-10-09 at 8 53 57

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -101,7 +101,7 @@ b {
 }
 
 .btn {
-    @apply h-full py-2 lt-md:px-2 md:px-4 rounded cursor-pointer;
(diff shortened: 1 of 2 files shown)


FireMasterK:
Thank you!

PR #1553, opened 2022-10-09
Update robots.txt (public instances)

shruuub:
For cleaner search results of public instances.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,2 +1,7 @@
 User-agent: *
-Disallow:
+Disallow: /channel/*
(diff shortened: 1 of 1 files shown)


FireMasterK:
Closing this with https://github.com/TeamPiped/Piped/commit/0c3d1f37f852fd7f5312817e17983a08eb2a627f instead, we now disallow all indexing!

shruuub:
Sounds good, also stops goofy crawlers (those who only sorta follow robots.txt, aka only some rules) from crawling :)!

PR #1572, opened 2022-10-15
Add proper whitespace to pinned comments

thecashewtrader:
Currently the `pinned_by` message doesn't have proper whitespace.

Before:
image

After:
image


FireMasterK:
Thanks for the pull request, but I ended up doing this differently.
Since some languages may require the uploader's name in front, I decided to add it to the translation string instead.
For example: https://github.com/TeamPiped/Piped/commit/67aaa00ff34c1de385473e7e0c194819273bca35#diff-531bff92eedfde00ed512bf7583e13ec2635140327e83ad536563da3ceba4625L83

Issue #842, opened 2022-02-19
Playback speed resets to default if right click screen


Yolakalemowa:
Been happening for a while but was too lazy to reproduce to know exactly how it happened...

Finally figured out it simply happens with you right click to play/pause.

M86xKC:
Found interesting behavior for the playback reset bug

steps to reproduce:
open any video
change playback speed to 2.0x
close that tab
open another video (in a new tab)
pause then play the video with mouse/touch
playback speed resets to 1.0x

I think what is happening here is that playback rate gets reset when [both conditions evaluates to false here somehow?](https://github.com/TeamPiped/Piped/blob/9d486118515ddefd9dd7d6cf9ca60ce156515704/src/main.js#L149) and playback rate is set to the default value as [in this line](https://github.com/TeamPiped/Piped/blob/b1efd97c40136cdc459c32eaf40135a6b33b6000/src/components/VideoPlayer.vue#L512)

PS:
- this bug does not get triggered by using space or k shortcuts.
- this bug was reproducible in both firefox & chrome.

M86xKC:
is this related to https://github.com/shaka-project/shaka-player/issues/3709?

FireMasterK:
Yes, it looks like it gets reset when using the UI :thinking:

FireMasterK:
Closing this with #1573.

Issue #725, opened 2022-01-13
Ability to set and save a default playback speed


Latrolage:

FireMasterK:
Buy changing it when playing a video retains it for other videos, so isn't this already implemented?

FireMasterK:
Could you let me know if https://rate-change-fixes.piped.pages.dev/ fixes the issue?

Latrolage:
After closing the browser window and coming back, it's set back to 1.0x speed again.

sdfg2:
The problem is nothing is 'saved' to the account other than the subscriptions.  Not even watched videos.

I like this idea, and I'd like to extend it to a per-channel setting - there are huge variations in how much some channels speed up the speech, I have to do this back and forth depending on what I'm watching.  But that would mean properly implementing user accounts.

FireMasterK:
Closing this with #1573.

PR #1573, opened 2022-10-16
Fix setting of initial playback rate

Zaczero:
Previously, *playbackRate* and *defaultPlaybackRate* were set on the `player` instead of the `videoEl`. Because of that, those settings were ignored and the `player.trickPlay` alone wasn't sufficient enough for the configuration to persist. After the fix, I removed the `player.trickPlay` call as it seemed redundant.

Related issues: #725, #842.


FireMasterK:
Thank you!

Issue #1512, opened 2022-09-30
Mark the video as watched.


Fertanima:
### Describe the feature

Option to mark video as watched.

### Why would this be useful to add?

Mark the video as watched.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

PR #1581, opened 2022-10-18
Feature: Mark as Watched button in Feed

crhallberg:
This PR adds a button next to the Listen and Add to Playlist buttons on feed.

- The button appears when Hide Watched is active.
- The PR also separates Hide Watched from Store Watch History, for users that want to only manually mark videos in their Feed.
- Right now this PR doesn't offer functionality to unmark a video as watched.

Addresses at least part of #1512.


src/components/VideoItem.vue
```diff
@@ -177,12 +184,48 @@ export default {
             const request = objectStore.get(this.video.url.substr(-11));
             request.onsuccess = event => {
                 const video = event.target.result;
-                if (video && (video.currentTime ?? 0) > video.duration * 0.9) {
+                if (
+                    video &&
+                    (!this.getPreferenceBoolean("watchHistory", false) ||
```
FireMasterK, line 189: This is redundant if you set the watched till as the duration of the video, which you do above.

FireMasterK (changes requested):

Bnyro:
Wouldn't it be better to use a favicon, e.g. an eye for the button instead of using plain text?

FireMasterK:
I feel we also need a new three-dot menu for playlist actions and this since otherwise, the UI is going to get cluttered.

src/components/VideoItem.vue
```diff
@@ -73,6 +73,13 @@
             >
                 <font-awesome-icon icon="circle-minus" />
             </button>
+            <button
+                v-if="isFeed && this.getPreferenceBoolean('hideWatched', false)"
+                @click="markAsWatched(video.url.substr(-11))"
```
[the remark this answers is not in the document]
crhallberg: That's not a bad idea. That can definitely be added.

src/components/PreferencesPage.vue
```diff
@@ -116,7 +116,7 @@
             @change="onChange($event)"
         />
     </label>
-    <label v-if="watchHistory" class="pref" for="chkHideWatched">
+    <label class="pref" for="chkHideWatched">
```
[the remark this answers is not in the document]
crhallberg: I'm definitely a use case where I don't want my watch history tracked automatically but really enjoy removing videos from my feed. I considered styling this as Archive or keeping the manually checked ones in a second database but that seemed ripe for issues.

src/components/VideoItem.vue
crhallberg: Thank you!

crhallberg:
> Wouldn't it be better to use an icon, e.g. an eye for the button instead of using plain text?

Not sure what you mean, this was styled with a Font Awesome eye icon to match the listen and add to playlist buttons.

Bnyro:
> Not sure what you mean, this was styled with a Font Awesome eye icon to match the listen and add to playlist buttons.
>
Oh, sorry, my bad, I just saw you added translation strings for it and just wondered if an icon wouldn't be better in that case, didn't see you already use one ;)

src/components/PreferencesPage.vue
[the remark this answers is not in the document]
FireMasterK: Okay, fair enough in that case!

src/components/VideoItem.vue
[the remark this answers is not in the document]
FireMasterK: Currently, it would just show "watched", which I think is a good start.

src/components/FeedPage.vue
```diff
@@ -85,6 +91,9 @@ export default {
                 this.loadMoreVideos();
             }
         },
+        onUpdateWatched() {
+            if (this.videos.length > 0) this.updateWatched(this.videos);
+        },
```
FireMasterK, line 96: We don't need to update all videos if just one video was watched!

crhallberg: I agree! However - as far as I can tell - `updateWatched` alters the array passed to it so `this.videos` needs to be used. Maybe there's something I can do here, constructing a smaller array with indexes or something, let me look into it.

crhallberg: Well that was easy lol

crhallberg:
Resolved the merge conflict. Looking forward to a new round of feedback.

Bnyro (approved):
I've quickly fixed some issues caused by merging which prevented it from working for me, looks good to me now!

crhallberg:
Apologies for the long hiatus! I have resolved the conflicts of this PR, hoping we can finish it off!

FireMasterK (changes requested):

crhallberg:
I still believe this feature is crucial to Piped but I do not have the bandwidth to continue development. Is there anyone who would like to take up the mantle?

PR #1608, opened 2022-10-22
Localize numbers using Internalisation API

### LINKED ISSUE #1299: Localize counts and units
### Describe the feature

Currently on the **homepage** you use units "k" (thousands) and "M" (million).

Truncating numbers up to a million with an abbreviated unit for thousands is not common in Croatian (and maybe other languages), and I don't think it is necessary.

IMO, everything up to one million should be shown on the **homepage** as a complete number (which is allready used when viewing an individual video).

Everything above one million could be truncated. In these cases a "space" should be added between the count and the unit, and the unit "M" should be translatable.

e.g.:
instead of 1.47M => 1.47 M (English)
instead of 1.47M => 1,47 mil. (Croatian, with Croatian abbreviation for million)

It improves the appearance of counts in a localized and typographically correct manner.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

thecashewtrader:
Closes #1299

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -88,20 +88,8 @@ const mixin = {
             return str;
         },
         numberFormat(num) {
-            const digits = 2;
-            const si = [
-                { value: 1, symbol: "" },
-                { value: 1e3, symbol: "k" },
-                { value: 1e6, symbol: "M" },
-                { value: 1e9, symbol: "B" },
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thank you!

PR #1659, opened 2022-11-02
Remember the share dialog settings

### LINKED ISSUE #1467: Remember settings in share Modal
### Describe the feature

It would be better if the modal remembers, what was selected at last like in selecting playlist.

This would be a great convenience feature. And also helpful in sharing YouTube links to people instead of Piped. As the Whole world is not using Piped and still some use youtube.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #1467

### DIFF
--- a/src/components/ShareModal.vue
+++ b/src/components/ShareModal.vue
@@ -66,8 +66,8 @@ export default {
             }
         },
         onChange() {
-            localStorage.setItem("shareWithTimeCode", this.withTimeCode);
-            localStorage.setItem("shareAsPipedLink", this.pipedLink);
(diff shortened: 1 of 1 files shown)


src/components/ShareModal.vue
```diff
@@ -63,6 +65,10 @@ export default {
                 alert(this.$t("info.cannot_copy"));
             }
         },
+        onChange() {
+            localStorage.setItem("shareWithTimeCode", this.withTimeCode);
+            localStorage.setItem("shareAsPipedLink", this.pipedLink);
```
FireMasterK, line 70: You should use the setPreference method, since this can fail if the user has cookies disabled.

Issue #803, opened 2022-02-08
Channel & Playlist View On Youtube Button


trymeouteh:
Please add a button on channel pages & playlist pages to view the channel/playlist on youtube.com

Loki-L1130:
Please elaborate about channel pages.

Playlist is currently being in the works. You can view a Playlist in Piped but you can play in a order or shuffle the order.

Loki-L1130:
I get it, now.

thecashewtrader:
I could implement this, blocked for the same reason as https://github.com/TeamPiped/Piped/issues/1611#issuecomment-1296262909

thecashewtrader:
@FireMasterK could you close this too? It doesn't seem to be closed automatically.

PR #1664, opened 2022-11-02
Add Watch on YouTube button

### LINKED ISSUE #1611: What Happen To The View On YouTube Button?
### Describe the feature

There used to be a "View On YouTube Button?" which when clicked on will take you to that video on youtube.com. Please bring this button back and please also add a button for [playlists and channels](https://github.com/TeamPiped/Piped/issues/803).

- To quickly go to the video on youtube.com.
- To quickly bookmark the video with the youtube.com URL and not the Piped instance URL.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

thecashewtrader:
Closes #1611 and #803

### DIFF
--- a/src/components/ChannelPage.vue
+++ b/src/components/ChannelPage.vue
@@ -35,10 +35,7 @@
             <font-awesome-icon icon="rss" />
         </a>
 
-        <WatchOnYouTubeButton
-            v-if="this.getPreferenceBoolean('showWatchOnYouTube', false)"
-            :link="`https://youtube.com/channel/${this.channel.id}`"
-        />
+        <WatchOnYouTubeButton :link="`https://youtube.com/channel/${this.channel.id}`" />
 
         <div class="flex mt-4 mb-2">
             <button
--- a/src/components/ChannelPage.vue
+++ b/src/components/ChannelPage.vue
@@ -35,6 +35,25 @@
             <font-awesome-icon icon="rss" />
         </a>
 
+        <!-- Watch on YouTube button: For large screens -->
+        <a
+            v-if="this.getPreferenceBoolean('showWatchOnYouTube', false)"
+            :href="`https://youtube.com/channel/${this.channel.id}`"
+            class="btn lt-lg:hidden ml-3"
+        >
+            <i18n-t keypath="player.watch_on" tag="strong">
+                <font-awesome-icon class="mx-1.5" :icon="['fab', 'youtube']" />
+            </i18n-t>
+        </a>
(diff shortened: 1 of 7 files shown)


Bnyro:
That's 3 times the same code:
A button that is rendered if the preference is enabled.

Creating a component here that's just used for the channel page, video page and playlist page would be better!

thecashewtrader:
Thanks for the feedback!

Bnyro:
The conditional statement should be part of the template as well preferably :)

thecashewtrader:
I considered it, but say for example in the future we add an on page toggle for the YouTube button like we have for Hide Recommendations/Comments (unlikely but just as an example), that integrated conditional could cause problems, also it looked a bit more clear this way. It's a semantic issue anyways, I could change it if you want.

Bnyro:
Hmm, I don't think we'll ever need to dynamically hide or show that button, @FireMasterK has the last word :wink:

IronMaltese:
I just wanna put my 2 cents in here

The follow link button should be replaced with the watch on YouTube button since there's no point in hitting follow link for a piped url

FireMasterK:
> there's no point in hitting follow link for a piped URL

Agree with this

IronMaltese:
One last thing: 

The share with time code button should be unchecked by default since I've never shared a video with a time code, and haven't seen anyone do it either

thecashewtrader:
How about this:
- The `v-if` goes into `WatchOnYouTubeButton`
- To just share/follow Piped link -> Copy from URL Bar
- To just share/follow YouTube link -> Use Watch On YouTube Button
- To share/follow Piped/YouTube link *with* timecode -> Use the share dialogue. Therefore, the share with timecode button is checked by default.

@IronMaltese thoughts?

Bnyro (approved):
LGTM!

Bnyro:
> To just share/follow Piped link -> Copy from URL Bar
> 
Following the Piped link doesn't really make sense currently, true, but think it might make sense as soon as we allow to choose a different Piped instance there so you can share links for the official instance easier while using a smaller or private instance for example.
>  To share/follow Piped/YouTube link _with_ timecode -> Use the share dialogue. Therefore, the share with timecode button is checked by default.
> 
I agree, we should enable sharing with timecode by default now as the dialog would be sorta useless otherwise.

thecashewtrader:
@Bnyro Sorry, I just copied the line so it was `share/follow` instead of just `share` lol

Issue #1672, opened 2022-11-03
Extremely long videos timeout/have an unhandled error


thecashewtrader:
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.kavin.rocks/) or is API related.

### Describe the bug

Extremely long videos (~8 days/200 hours) do not play.

### To Reproduce

Open this video: https://piped.kavin.rocks/watch?v=mivcDoTjpQE

### Expected behavior

The video should play as normal

### Logs/Errors

```
Uncaught (in promise) 
Object { severity: 2, category: 7, code: 7000, data: [], handled: false }
​
{
  "severity": 2,
  "category": 7,
  "code": 7000,
  "data": [],
  "handled": false
}
```

### Browser, and OS with Version.

Mozilla Firefox 106.0.4 and Chromium 107.0.5304.87 on Arch Linux btw

### Additional context

It seems that YouTube changed it's policy to allow a max length of 12 hours for videos, but videos uploaded before the change would stay up. Maybe this has something to do with the problem?

PR #1673, opened 2022-11-03
Allow timecode to be used as timestamp in Share Modal

thecashewtrader:
Currently, the share modal only allows timestamp to be in seconds. This PR allows the timestamp to be in seconds as well as in `HH:MM:SS` format.

Some points to note:
- Currently, YouTube has a bit confusing of a policy about maximum video duration, as stated in #1672. Therefore, this PR parses the timecode string manually instead of using something like `Intl.DateFormat`.
- Currently, in the `HH:MM:SS` format, the seconds/minutes/hours value can be more than expected, which is handled without errors. For example: `10:61` will return `&t=661`. This can be changed if intended, but imo it's quite convenient.
- It would be required to update the `actions.time_code` translations to reflect the new changes.

### DIFF
--- a/src/components/ShareModal.vue
+++ b/src/components/ShareModal.vue
@@ -70,24 +70,29 @@ export default {
             timeStamp: null,
             hasPlaylist: false,
             showQrCode: false,
+            durations: [1, 60, 60 * 60, 60 * 60 * 24],
         };
     },
     computed: {
         generatedLink() {
-            var baseUrl = this.pipedLink
+            const baseUrl = this.pipedLink
                 ? window.location.origin + "/watch?v=" + this.videoId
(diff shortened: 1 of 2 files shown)


FireMasterK:
I fail to understand what's wrong with the current mechanism, is the number overflowing? (I highly doubt that's the case)

thecashewtrader:
> I fail to understand what's wrong with the current mechanism, is the number overflowing? (I highly doubt that's the case)

imo it's more convenient to put in a `HH:MM:SS` string than a raw number, as you don't have to do the calculation yourself, especially if it's a long video.

Bnyro (approved):
Rebased and cleaned up a bit, thanks!

PR #1675, opened 2022-11-03
Make numberFormat respect language and region settings

thecashewtrader:
Currently,the numberFormat function will *always* use the fallback locale provided by Intl.NumberFormat. This PR tries to use the user's preferred language and region. If that fails, it will default to the fallback locale.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -88,7 +88,12 @@ const mixin = {
             return str;
         },
         numberFormat(num) {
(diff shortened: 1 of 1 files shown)


FireMasterK:
Thanks!

thecashewtrader:
Oh sorry, I thought the matcher would automatically fallback, my bad.

PR #1682, opened 2022-11-05
Add Hong Kong region to CountryMaps

### LINKED ISSUE #1678: No Hong Kong option in the Country Selection
### Describe your question

Preferences--Country Selection
No Hong Kong (china) option, the trending is nothing if choose China but YouTube has it.
There are very many Chinese people using this great project, including people like me who speak poor English, could you please add the Hong Kong option?

AlyoshaVasilieva:
Fixes #1678. Hong Kong isn't a country, so it seems it wasn't included in the CountryMaps; however, it is a region that YouTube operates in, while mainland China isn't. This allows Hong Kong to be selected as a country for things such as the trending page.

For most of the languages, I used EU official data: https://publications.europa.eu/code/en/en-5000500.htm (language selector at top right)  
For Serbian and Chinese Traditional I looked it up online. Hopefully I got it right.

I tried to keep the addition alphabetically sorted according to the name. The Serbian and Chinese maps seem to match English order, though.


FireMasterK:
I am not a fan of this as a permanent solution. In the future, I want to maintain a list of supported country codes and then do something like this: https://stackoverflow.com/a/65787047

PR #1692, opened 2022-11-07
Add better error messages for localStorage disabled

FireMasterK:


Bnyro:
Wouldn't the alert now even pop up if the user checks an option in the share dialog for example. That would be kind of misleading in my opinion as sharing a video with a time stamp for example doesn't require the user to have the local storage enabled, only the preference does. 
Maybe we should introduce an optional parameter `required` to the `setPreferenceItem()` function which is handled as true by default.
When it's passed as false (which would be used for things that are just saved automatically at user interactions but are not required to be kept stored, e.g. settings in the share dialog), the alert dialog wouldn't show and thus it wouldn't give some "false alerts" in some cases.
Only a suggestion though, how it is currently would be fine as well, just wouldn't cover such cases properly in my opinion.

PR #1697, opened 2022-11-08
Fix Add to playlist button margin

thecashewtrader:
Currently, the margins/gaps are different:
image

After:

image


Bnyro:
Thanks!

Issue #1391, opened 2022-09-05
Redesign - EFY


dragos-efy:
Can be optimised more obviously, but here's the new Piped redesign! 🥳

For people that are not part of the matrix room, it uses [efy ui](https://github.com/dragos-efy/efy) to add more customisation features, so things changed a bit, but I'm working on the documentation to make it easier for devs to understand the logic behind it and contribute. Any feedback or suggestions you have, please share them, I and other really cool people who I'm grateful for have been testing it, but bugs can still come up, therefore if you wanna share something, please do!

Live demo: https://efy.piped.pages.dev

Special thanks to: Bnyro, Kavin, Zucc (whatever.social), Diamond_AaronXG and other small or anonymous contributions, you guys rock! 🔥

Bnyro:
The code for it can be found at the `efy` branch if anyone is willing to test it locally or do anything else with the code :)

Senor-Ducky:
Bug: The search bar appears to be shrunk and not working right.
![image](https://user-images.githubusercontent.com/61725851/188673018-f755b3f9-7184-4019-a9ac-95dd937f8a1d.png)

dragos-efy:
@Senor-Ducky It got broken after the last commit, it was okay before, but I think I identified the cause for the most part. Thank you! If you notice other bugs lmk

dragos-efy:
The live instance will likely have issues for a while, but the css code is fine, it's just needs to be split and distributed to vue components. Keep sending feedback, but keep in mind that there might be a difference between how unstable it seems to be and how stable it actually is

IronMaltese:
![image](https://user-images.githubusercontent.com/89056889/188745034-e5e8fafb-1a09-488c-982a-879b71a5f696.jpeg)

The arrow seems to be too close to the word

dragos-efy:
@IronMaltese Glad you like it! Your issue was on the list too, but having it here makes it easier to keep track of what I gotta do lol, so thanks!

dragos-efy:
Everyone and @Senor-Ducky , if you tried the live instance before the problem got fixed, try clearing your cache (`CTRL+F5` on some browsers) or opening it in a private tab.  And remember, if you need to clear, import or export your local storage or background images, go to `Save & Restore` in the efy menu. Things should be as stable as they were before + a few small issues that I'll be working on when I have time. Planning to fix all the issues mentioned here in the next 2-3 days or sooner. Enjoy & keep giving feedback! 🥳

[...]

PR #1720, opened 2022-11-10
Improve playlist page when not logged in

Bnyro:
2022-11-10T18:59:07,700586231+01:00

see https://github.com/TeamPiped/Piped/issues/1391#issuecomment-1309360266

That page appears when the user navigates to the playlists page while not being logged in as recent changes of @dragos-efy added the previous redirect resulted in a dialog that couldn't be dismissed.

### DIFF
--- a/src/components/PlaylistsPage.vue
+++ b/src/components/PlaylistsPage.vue
@@ -3,39 +3,67 @@
 
     <hr />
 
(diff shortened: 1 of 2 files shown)


FireMasterK (approved):
Nice, I like this approach over the previous one!

dragos-efy:
same, it's cool!

PR #1756, opened 2022-11-15
Don't load trending page on feed homepage

photochlorination:
This prevents the trending data from being fetched unnecessarily before redirecting to the feed.


FireMasterK:
Thanks!

PR #1768, opened 2022-11-17
Hide sponsor options from Preferences if sponsorBlock isn't checked

bonswouar:
Tiny hack but I don't really now VueJS so don't skip the review please :smile:


FireMasterK:
Thanks!

PR #1778, opened 2022-11-18
add support for auto deleting the history

### LINKED ISSUE #1759: Add function to delete history older than x days/weeks/months
### Describe the feature

On the Piped /history page I would like to have the ability to delete videos that are **_older_** than x days/weeks/months. 
Right now you can only delete ALL history. 

It would also be nice if you could delete select videos from the history. 

I like to keep _some_ viewing history, e.g. the last month, but I don't want it to keep it for months/years and I also don't want to delete all history. 

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #1759

### DIFF
--- a/src/components/HistoryPage.vue
+++ b/src/components/HistoryPage.vue
@@ -1,13 +1,32 @@
 <template>
-    <h1 class="font-bold text-center" v-t="'titles.history'" />
+    <h1 class="font-bold text-center mb-3" v-t="'titles.history'" />
 
-    <div class="flex md:items-center gap-2 flex-col md:flex-row">
-        <button class="btn" v-t="'actions.clear_history'" @click="clearHistory" />
+    <div class="flex">
+        <div class="flex md:items-center gap-2 flex-col md:flex-row">
+            <button class="btn" v-t="'actions.clear_history'" @click="clearHistory" />
 
-        <button class="btn" v-t="'actions.export_to_json'" @click="exportHistory" />
+            <button class="btn" v-t="'actions.export_to_json'" @click="exportHistory" />
 
-        <div class="ml-auto flex gap-1 items-center">
-            <SortingSelector by-key="watchedAt" @apply="order => videos.sort(order)" />
+            <div class="ml-auto flex gap-1 items-center">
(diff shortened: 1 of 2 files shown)


src/components/HistoryPage.vue
```diff
@@ -32,27 +41,37 @@ export default {
...
+                                uploaderName: video.uploaderName,
+                                uploaderUrl: video.uploaderUrl,
+                                duration: video.duration,
+                                thumbnail: video.thumbnail,
+                                watchedAt: video.watchedAt,
+                            });
+                        } else {
+                            store.delete(video.videoId);
+                        }
                         if (this.videos.length < 1000) cursor.continue();
```
FireMasterK, line 75: This line should be moved to before we push items.

FireMasterK (approved):

FireMasterK (changes requested):
Actually, we should use an index instead for removing data: https://stackoverflow.com/a/15625231

The lower index start would be `0` and the upper time would be the max time difference.

We could run this as an `async` function.

src/components/HistoryPage.vue
```diff
@@ -71,6 +90,28 @@ export default {
             }
             this.videos = [];
         },
+        onChange() {
+            this.setPreference("autoDeleteWatchHistory", this.autoDeleteHistory);
+            this.setPreference("autoDeleteWatchHistoryDelay", this.autoDeleteDelay);
+            this.setPreference("autoDeleteWatchHistoryTimeUnit", this.autoDeleteTimeUnit);
```
FireMasterK, line 96: Can we have this in one variable? (Maybe store it as seconds or milliseconds, and convert it?)

Bnyro: Wouldn't this cause some issues UX-wise?
For example if the user sets the delay to 36 hours, we would probably then show `days` as the selection because the time is gt 24 hours. We could also use some modulos, but that would overcomplicate things in my opinion. Would speaks against storing two different variables in order to keep the code easier to read (in my opinion)?

PR #1784, opened 2022-11-19
Fix Minimize Comments preferences toggle

cocoanaut:
Fixes the "Minimize Comments by default" toggle in Preferences.


Bnyro:
Thanks!

PR #1826, opened 2022-11-27
Adds Taiwan as a YouTube Location

bruvduroiu:
YouTube has support for, and operates in Taiwan.

Adding this into Piped also allows viewing Trending content from Taiwan.

I've used:
- https://publications.europa.eu/code/en/en-5000500.htm for most of the European languages
- Google Translate for Serbian

I've tried to sort the `name` field, and insert based on UTF-8 codes, however seems like `sr` and `zh_Hant` maps are not ordered properly.


FireMasterK:
Thanks!

PR #1831, opened 2022-11-28
Playlists import and export

### LINKED ISSUE #1018: Import/Export Playlist
We now have the ability to create playlists in Piped (#358)... it would be awesome to have the ability to import a playlist from a YouTube link (as they will of course have links containing the IDs of the vids) and also hopefully provide the ability to export playlists (for at least to other instances of Piped/derivatives and ideally for NewPipe as well).

I know I've got a playlist of over 1700 videos I've collated over at least 5 years and I don't want to have to recreate it when migrating away from goggle entirely, I doubt I'll be alone (previously mentioned by @trymeouteh in aforementioned issue)...

Bnyro:
closes #1018 
compatible with the Invidious settings export
also see https://github.com/iv-org/invidious/issues/2897

### DIFF
--- a/src/components/PlaylistsPage.vue
+++ b/src/components/PlaylistsPage.vue
@@ -4,10 +4,16 @@
     <hr />
 
     <div class="flex justify-between mb-3">
-        <button v-t="'actions.create_playlist'" class="btn" @click="createPlaylist" />
+        <button v-t="'actions.create_playlist'" class="btn" @click="onCreatePlaylist" />
         <div class="flex">
-            <button v-t="'actions.export_to_json'" class="btn" @click="exportPlaylists" />
-            <button v-t="'actions.import_from_json'" class="btn ml-2" @click="importPlaylists" />
+            <button
+                v-if="this.playlists.length > 0"
+                v-t="'actions.export_to_json'"
+                class="btn"
+                @click="exportPlaylists"
+            />
+            <input id="fileSelector" ref="fileSelector" type="file" class="display-none" @change="importPlaylists" />
+            <label for="fileSelector" v-t="'actions.import_from_json'" class="btn ml-2" />
         </div>
     </div>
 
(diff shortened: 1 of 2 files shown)


src/components/PlaylistsPage.vue
```diff
@@ -94,22 +106,96 @@ export default {
...
+                        if (playlists.length != importedCount) return;
+                        window.location.reload();
+                    });
+                }
+            });
+        },
+        async createPlaylistWithVideos(playlist, onSuccess) {
+            let newPlaylist = await this.createPlaylist(playlist.name);
+            console.log(newPlaylist);
+            let videoIds = playlist.videos.map(url => url.replace("https://youtube.com/watch?v=", ""));
```
FireMasterK, line 183: You should use substr(-11) instead :)

src/components/PlaylistsPage.vue
```diff
@@ -94,22 +106,96 @@ export default {
...
+            const file = this.$refs.fileSelector.files[0];
+            file.text().then(text => {
+                let playlists = JSON.parse(text).playlists;
+                if (!playlists.length) {
+                    alert(this.$t("actions.no_valid_playlists"));
+                    return;
+                }
+                let importedCount = 0;
+                for (var i = 0; i < playlists.length; i++) {
+                    this.createPlaylistWithVideos(playlists[i], () => {
```
FireMasterK, line 172: Instead of doing this, you should add all the promises to an array, and then call `Promise.all()`

src/components/PlaylistsPage.vue
```diff
@@ -94,22 +106,96 @@ export default {
...
+                type: "playlist",
+                // as Invidious supports public and private playlists
+                visibility: "private",
+                // list of the videos, starting with "https://youtube.com" to clarify that those are YT videos
+                videos: [],
+            };
+            for (var i = 0; i < playlist.relatedStreams.length; i++) {
+                playlistJson.videos.push("https://youtube.com" + playlist.relatedStreams[i].url);
+            }
+            onSuccess(playlistJson);
```
FireMasterK, line 160: Rather than this, return the data and then handle it.

src/components/PlaylistsPage.vue
```diff
@@ -94,22 +106,96 @@ export default {
...
+        exportPlaylists() {
+            if (!this.playlists) return;
+            let json = {
+                format: "Piped",
+                version: 1,
+                playlists: [],
+            };
+            let playlistsSize = this.playlists.length;
+            for (var i = 0; i < playlistsSize; i++) {
+                this.fetchPlaylistJson(this.playlists[i].id, playlist => {
```
FireMasterK, line 139: This should also be replaced with `Promise.all()`

FireMasterK (changes requested):

FireMasterK (approved):
Wow, incredible!

PR #1890, opened 2022-12-15
Update - 2022.12.15 Beta

dragos-efy:
- translations and dynamic translation logic
- custom gaps and gradient angle
- new gradient logic
- intuitive background images
- more integrated server support
- keyboard accessibility
- trans filters cover more elements
- no more white flash on dark mode
- speed optimisation
- more icons
- bug fixes


FireMasterK:
Question - EFY is currently a git submodule, why are we adding it again?

Bnyro:
> Question - EFY is currently a git submodule, why are we adding it again?
>
That's what I discussed with Dragos in PM, we probably should have done that here. We'll need to revert that change before merging.

dragos-efy:
@FireMasterK yeah, I made it work yesterday with just the efy submodule inside the public folder yesterday, so if it goes well we can try that. 2 folders is clearly annoying to maintain and it increases the size for no reason.

As for why I was doing the weird shenaningan move, it's cuz external files were not loading properly.

PR #1951, opened 2023-01-06
Playlist bookmarks

### LINKED ISSUE #1742: Bookmark playlists (saving without cloning)
### Describe the feature

Piped can store playlists without cloning them. 

I have to save some playlists that get continuously updated. So It's kind of annoying to delete and clone everytime. Is there any chance of adding this feature in near future?

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #1742

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -40,12 +40,13 @@ export default {
         darkModePreference.addEventListener("change", () => {
             this.setTheme();
         });
-        if (this.getPreferenceBoolean("watchHistory", false))
-            if ("indexedDB" in window) {
-                const request = indexedDB.open("piped-db", 2);
-                request.onupgradeneeded = ev => {
-                    const db = request.result;
-                    console.log("Upgrading object store.");
+
+        if ("indexedDB" in window) {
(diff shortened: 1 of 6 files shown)


src/App.vue
```diff
@@ -40,12 +40,13 @@ export default {
...
-                request.onupgradeneeded = ev => {
-                    const db = request.result;
-                    console.log("Upgrading object store.");
+
+        if ("indexedDB" in window) {
+            const request = indexedDB.open("piped-db", 3);
+            request.onupgradeneeded = ev => {
+                const db = request.result;
+                console.log("Upgrading object store.");
+                if (this.getPreferenceBoolean("watchHistory", false)) {
```
FireMasterK, line 49: I feel we should create the store even if it not enabled, since this code is essentially ran once since its a migration.

FireMasterK:
Looks like a side-effect of this PR is, we essentially enable watch history for everyone, even when we don't have it enabled!
Should be easy enough to fix tho!

PR #2002, opened 2023-01-18
Make timestamps in comments seek instead of reload

Tyrritt:
Small quality of life improvement. This will make clicking timestamps posted as `HH:MM:SS` (or `MM:SS`) in the comments seek the video to the correct place instead of reloading the page with a `&t=ssss` parameter, similar to the chapter bar. The links can still be copied via right click and opened in a new tab as usual.

Example video without chapters: https://piped.video/watch?v=cVFzblT5VPE, see tracklist comment.

### DIFF
--- a/src/components/CommentItem.vue
+++ b/src/components/CommentItem.vue
@@ -71,6 +71,7 @@ export default {
         uploader: { type: String, default: null },
         videoId: { type: String, default: null },
     },
+    emits: ["seek"],
     data() {
         return {
             loadingReplies: false,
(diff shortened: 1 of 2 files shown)


src/components/CommentItem.vue
```diff
@@ -79,6 +80,21 @@ export default {
...
+                elem.addEventListener("click", function (event) {
+                    if (!event || !event.target) return;
+                    if (!event.target.getAttribute("href").match(/(?<=t=)\d{1,}/)) return;
+
+                    const time = parseInt(event.target.getAttribute("href").match(/(?<=t=)\d{1,}/)[0]);
+                    thisComment.$emit("seek", time);
+                    event.preventDefault();
+                });
+            }
+        });
```
FireMasterK, line 96: Why can't we add just one event listener in the document/window and check if the element matches? (Through the concept of Event Bubbling)

Tyrritt:
I hadn't thought of letting the event bubble up and have just one EventListener catch the relevant ones. 
Update to implement the suggestion and remove the for-each loop in CommentItem

FireMasterK:
You should also remove the listener when the component unmounts (to prevent memory leaks / other potential issues).

FireMasterK (approved):

FireMasterK:
Thank you!

PR #2022, opened 2023-01-26
Allow channel tabs as query parameter

### LINKED ISSUE #2011: Remember which "tab" of the channel page is currently selected
### Describe the feature

If the user visits a channel page, and clicks "Playlists" for example, the state of the UI changes correctly, but that state is lost if the user refreshes the page in their browser.

Similarly, this prevents a user from giving someone else a link to a specific view of a channel page.

The currently selected "tab" could be stored as a URL parameter or URL fragment, or you could do something clever with cookies if you don't want to support deep linking to non-default views of the page for some reason.

I should also point out that, in the context of other page refresh bugs (like #1906) this means in practice that the UI state keeps getting reset, against the user's wishes, so implementing this "feature request" would greatly reduce the impact of that bug.

Even if that other bug were fixed, I think that it is bad UX to make the user have to re-select the tab they want after refreshing the page, and I think it is good UX to allow deep linking into views of web apps.  Another example of the benefit of deep linking is that it allows specific views to be bookmarked, which makes navigating to that view require half as many actions.

### Concept(s)

_No response_

### Additional context

I'm really grateful for the hard work that has gone into this project, and thank you for taking the time to consider this feature request.

### Acknowledgements

Bnyro:
closes #2011

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -251,18 +251,6 @@ const mixin = {
             elem.click();
             elem.remove();
         },
-        appendQueryToUrl(key, value) {
-            let searchParams = new URLSearchParams(window.location.search);
-            searchParams.set(key, value);
-            let newurl =
-                window.location.protocol +
-                "//" +
-                window.location.host +
-                window.location.pathname +
-                "?" +
-                searchParams.toString();
-            window.history.pushState({ path: newurl }, "", newurl);
-        },
     },
     computed: {
         authenticated(_this) {
--- a/src/main.js
+++ b/src/main.js
@@ -251,6 +251,18 @@ const mixin = {
             elem.click();
             elem.remove();
         },
+        appendQueryToUrl(key, value) {
+            let searchParams = new URLSearchParams(window.location.search);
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -251,6 +251,18 @@ const mixin = {
...
+            let searchParams = new URLSearchParams(window.location.search);
+            searchParams.set(key, value);
+            let newurl =
+                window.location.protocol +
+                "//" +
+                window.location.host +
+                window.location.pathname +
+                "?" +
+                searchParams.toString();
+            window.history.pushState({ path: newurl }, "", newurl);
```
FireMasterK, line 264: Rather than pushState, which creates a history entry in the browser, why not replaceState? I think this would result in a cleaner browser history.

Todo to self: Refactor all these helper functions to JS files

FireMasterK: I also noticed that this can be done a lot simpler: https://developer.mozilla.org/en-US/docs/Web/API/History/pushState#change_a_query_parameter

PR #2026, opened 2023-01-27
Add `shorts` and `videos` filter to the feed

### LINKED ISSUE #1578: Option to remove shorts from feed
### Describe the feature

Shorts is the worst thing to look at on my feed. Being able to disable it would be a great option.

Helps clean up the subscriptions page to just published videos and not videos labeled "Shorts"

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #1578

### DIFF
--- a/src/components/FeedPage.vue
+++ b/src/components/FeedPage.vue
@@ -99,13 +99,13 @@ export default {
             }
         },
         shouldShowVideo(video) {
-            switch (this.selectedFilter) {
-                case "all":
-                    return true;
+            switch (this.selectedFilter.toLowerCase()) {
+                case "shorts":
+                    return video.isShort;
                 case "videos":
                     return !video.isShort;
                 default:
-                    return video.isShort;
+                    return true;
             }
         },
         onFilterChange() {
--- a/src/components/FeedPage.vue
+++ b/src/components/FeedPage.vue
@@ -11,6 +11,13 @@
         </a>
     </span>
 
+    <label for="filters" class="ml-10 mr-2">
+        <strong v-text="`${$t('actions.filter')}:`" />
+    </label>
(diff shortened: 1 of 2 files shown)


src/components/FeedPage.vue
```diff
@@ -51,6 +62,8 @@ export default {
             this.loadMoreVideos();
             this.updateWatched(this.videos);
         });
+
+        this.selectedFilter = this.$route.query.filter ?? this.getPreferenceString("feedFilter") ?? "all";
```
FireMasterK, line 66: `this.$route.query.filter` is technically not needed, since `this.getPreferenceString("feedFilter")` checks for the parameter in the query.

src/components/FeedPage.vue
```diff
@@ -85,6 +98,19 @@ export default {
...
             }
         },
+        shouldShowVideo(video) {
+            switch (this.selectedFilter) {
+                case "all":
+                    return true;
+                case "videos":
+                    return !video.isShort;
+                default:
+                    return video.isShort;
```
FireMasterK, line 108: I'd keep it like this:
```
                case "shorts":
                    return video.isShort;
                case "videos":
                    return !video.isShort;
                default:
                    return true;
```

We could also convert the query string/preference to lowercase too above.

src/components/FeedPage.vue
Bnyro: Adding that was intended to also allow customizing the filter with a query parameter since using queries instead of preferences seems to be quite popular among the users of Piped

FireMasterK: The getPreference* methods check the query string first, then followed by localstorage. Isn't that the same thing?

Bnyro: Oh, you're right, I misunderstood you before :)

PR #2055, opened 2023-02-01
Add more options for sponsorblock

### LINKED ISSUE #894: Feature Request: SponsorBlock add prompt option
Currently Piped has the option to skip or none for sponsorblock. It would be great to have a prompt option as well which will display a skip button instead of the current automatic skipping.

This is similar functionality to viewtube.io

The idea behind it is sometimes you don't always want to skip, so this would but the skipping into the hands of the user.

9ary:
This overhauls the skipping logic and adds an option to show a skip button instead of autoskipping (configurable per category). The enter key also serves as a hotkey to skip. Fixes #894.

I've also added a simple option to prune segments shorter than a certain duration.

Known issues/possible future improvements:
- ~The skip button is ugly. Advice on making it look good would be appreciated, I'm not a web person.~
- ~The button location is suboptimal to be honest, it'd be nice to have a minimally intrusive overlay that doesn't get hidden.~
- It might be helpful to show the segment type on the skip button, but exisiting localization strings didn't look like they were flexible enough for that.
- An "unskip" action might be nice to have.

### DIFF
--- a/src/components/PreferencesPage.vue
+++ b/src/components/PreferencesPage.vue
@@ -184,83 +184,77 @@
         />
     </label>
     <div v-if="sponsorBlock">
-        <label class="pref" for="chkSkipSponsors">
+        <label class="pref" for="ddlSkipSponsors">
             <strong v-t="'actions.skip_sponsors'" />
-            <input
-                id="chkSkipSponsors"
-                v-model="skipSponsor"
-                class="checkbox"
-                type="checkbox"
-                @change="onChange($event)"
-            />
+            <select id="ddlSkipSponsors" v-model="skipSponsor" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
+                <option v-t="'actions.skip_button_only'" value="button" />
+                <option v-t="'actions.skip_automatically'" value="auto" />
+            </select>
         </label>
-        <label class="pref" for="chkSkipIntro">
+        <label class="pref" for="ddlSkipIntro">
             <strong v-t="'actions.skip_intro'" />
-            <input id="chkSkipIntro" v-model="skipIntro" class="checkbox" type="checkbox" @change="onChange($event)" />
+            <select id="ddlSkipIntro" v-model="skipIntro" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
+                <option v-t="'actions.skip_button_only'" value="button" />
+                <option v-t="'actions.skip_automatically'" value="auto" />
+            </select>
         </label>
-        <label class="pref" for="chkSkipOutro">
+        <label class="pref" for="ddlSkipOutro">
             <strong v-t="'actions.skip_outro'" />
-            <input id="chkSkipOutro" v-model="skipOutro" class="checkbox" type="checkbox" @change="onChange($event)" />
+            <select id="ddlSkipOutro" v-model="skipOutro" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
+                <option v-t="'actions.skip_button_only'" value="button" />
+                <option v-t="'actions.skip_automatically'" value="auto" />
+            </select>
         </label>
-        <label class="pref" for="chkSkipPreview">
+        <label class="pref" for="ddlSkipPreview">
             <strong v-t="'actions.skip_preview'" />
-            <input
-                id="chkSkipPreview"
-                v-model="skipPreview"
-                class="checkbox"
-                type="checkbox"
-                @change="onChange($event)"
-            />
+            <select id="ddlSkipPreview" v-model="skipPreview" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
+                <option v-t="'actions.skip_button_only'" value="button" />
+                <option v-t="'actions.skip_automatically'" value="auto" />
+            </select>
         </label>
-        <label class="pref" for="chkSkipInteraction">
+        <label class="pref" for="ddlSkipInteraction">
             <strong v-t="'actions.skip_interaction'" />
-            <input
-                id="chkSkipInteraction"
-                v-model="skipInteraction"
-                class="checkbox"
-                type="checkbox"
-                @change="onChange($event)"
-            />
+            <select id="ddlSkipInteraction" v-model="skipInteraction" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
+                <option v-t="'actions.skip_button_only'" value="button" />
+                <option v-t="'actions.skip_automatically'" value="auto" />
+            </select>
         </label>
-        <label class="pref" for="chkSkipSelfPromo">
+        <label class="pref" for="ddlSkipSelfPromo">
             <strong v-t="'actions.skip_self_promo'" />
-            <input
-                id="chkSkipSelfPromo"
-                v-model="skipSelfPromo"
-                class="checkbox"
-                type="checkbox"
-                @change="onChange($event)"
-            />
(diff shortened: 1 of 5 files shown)


src/components/PreferencesPage.vue
```diff
@@ -497,7 +502,18 @@ export default {
...
+                if (skipOptions.sponsor !== undefined) this.skipSponsor = skipOptions.sponsor;
+                if (skipOptions.intro !== undefined) this.skipIntro = skipOptions.intro;
+                if (skipOptions.outro !== undefined) this.skipOutro = skipOptions.outro;
+                if (skipOptions.preview !== undefined) this.skipPreview = skipOptions.preview;
+                if (skipOptions.interaction !== undefined) this.skipInteraction = skipOptions.interaction;
+                if (skipOptions.selfpromo !== undefined) this.skipSelfPromo = skipOptions.selfpromo;
+                if (skipOptions.music_offtopic !== undefined) this.skipMusicOffTopic = skipOptions.music_offtopic;
+                if (skipOptions.poi_highlight !== undefined) this.skipHighlight = skipOptions.poi_highlight;
+                if (skipOptions.filler !== undefined) this.skipFiller = skipOptions.filler;
+            } else if (localStorage.getItem("selectedSkip") !== null) {
```
FireMasterK, line 516: Please use `getPreferenceString` here.

src/components/PreferencesPage.vue
```diff
@@ -546,6 +562,7 @@ export default {
             }
 
             this.showMarkers = this.getPreferenceBoolean("showMarkers", true);
+            this.minSegmentLength = Math.max(Number(localStorage.getItem("minSegmentLength")), 0);
```
FireMasterK, line 565: Use the `getPreferenceNumber` method here.

src/components/WatchVideo.vue
```diff
@@ -367,15 +367,29 @@ export default {
...
-                    this.getPreferenceString("selectedSkip", "sponsor,interaction,selfpromo,music_offtopic").replaceAll(
-                        ",",
-                        '","',
-                    ) +
-                    '"]',
+            var selectedSkip = this.getPreferenceString(
+                "selectedSkip",
+                "sponsor,interaction,selfpromo,music_offtopic",
+            ).split(",");
+            const skipOptionsJSON = localStorage.getItem("skipOptions");
```
FireMasterK, line 374: Please use a `getPreference` method here.

src/components/PreferencesPage.vue
9ary: Sure thing, this is based on existing code and I wasn't sure why there were two different ways to read preferences.

src/components/WatchVideo.vue
9ary: In this case, I'm going to add a `getPreferencesJSON()` to support loading complex preferences from storage. A matching setter also feels appropriate.

src/components/VideoPlayer.vue
```diff
@@ -458,6 +519,17 @@ export default {
                 }
 
                 const config = {
+                    controlPanelElements: [
+                        "play_pause",
+                        "time_and_duration",
+                        "spacer",
+                        "skip_segment",
```
FireMasterK, line 526: When SponsorBlock is not enabled, or I use the auto skip mode, wouldn't this be unnecessary?

9ary: The button is hidden by default, it's only visible when we're in a segment, so I didn't feel that gating it further was necessary.

9ary:
It's been a few days. Are there any blockers I can address from my end, or is it just lack of time to review? (no pressure!)

FireMasterK:
Will try to get this reviewed in a few hours today :)

src/components/PreferencesPage.vue
```diff
@@ -184,83 +184,77 @@
...
             <strong v-t="'actions.skip_sponsors'" />
-            <input
-                id="chkSkipSponsors"
-                v-model="skipSponsor"
-                class="checkbox"
-                type="checkbox"
-                @change="onChange($event)"
-            />
+            <select id="ddlSkipSponsors" v-model="skipSponsor" class="select w-auto" @change="onChange($event)">
+                <option v-t="'actions.no'" value="no" />
```
Bnyro, line 190: To keep it easier to overview, maybe we should create an array containing the possible options and then use
```vue
<option v-for="option in sbOptions" v-t=`actions.skip_sponsor_${option}` :value="option">`
```
which would save us some repeating lines and allow extending it in the future (if we wanted to)

FireMasterK: True, we could potentially even maybe create a component for this.

Bnyro: Yeah, thought about that too but I wasn't sure whether we can really make that possible in a good way as `v-model` requires us to be able to access the variable directly which we can't really do when using a component

Bnyro: Though Vue might have a way for it I'm not familiar with, but I've never heard of pointers in JS :upside_down_face:

9ary: I think maybe putting v-for on the <label> element would be smarter, that way we could avoid all that repetition altogether?

Bnyro: But how should we use the booleans like `skipIntro` as v-model then?

9ary: Good question. I'm not familiar with vue at all, but let's see what I can come up with.

FireMasterK: https://vuejs.org/guide/components/v-model.html seems to be helpful for this.

9ary: Latest commit should do it.

Bnyro (approved):

9ary:
Having worked on the SB code here are a few small things I've noticed and will leave up for grabs because I don't care enough to work on them myself:
- The "highlight" segment type is mishandled: it shouldn't be skipped _over_ but skipped _to_. The role of this segment is to quickly get to the point of the video. The SB extension gives the user a choice of automatically skipping when the video starts, or showing a prompt. This probably should handle `?t=` in the URL.
- We probably need to handle overlapping segments. The extension will show multiple prompts for those, I think that's overkill. But right now if a segment that's set to button only is followed by an overlapping autoskip segment, the autoskip will only happen after the first segment ends.
- Since the skip button is intrusive, some users may want a "show in seekbar only" option. This is configurable per segment type in the SB extension.

These should all be easy to implement considering the groundwork laid down by this PR.

src/components/PreferencesPage.vue
```diff
@@ -497,55 +440,25 @@ export default {
...
+            var skipOptions, skipList;
+            if ((skipOptions = this.getPreferenceJSON("skipOptions")) !== undefined) {
+                Object.entries(skipOptions).forEach(([key, value]) => {
+                    var opt = this.skipOptions.get(key);
+                    if (opt !== undefined) opt.value = value;
+                    else console.log("Unknown sponsor type: " + key);
+                });
+            } else if ((skipList = this.getPreferenceString("selectedSkip")) !== undefined) {
+                skipList = skipList.split(",");
+                this.skipOptions.forEach(opt => (opt.value = "no"));
```
FireMasterK, line 452: The problem I see in this, is if we add new categories / preferences for sponsorblock going forward, we're not handling that in any way.

9ary: This is a problem with the old config format and the new one solves this, I'm only handling selectedSkip for backwards compatibility. Maybe migrating more aggressively would be better?
To be honest I don't think we should worry about this for now, it's easy enough to fix when it becomes necessary.

FireMasterK (approved):

FireMasterK:
Thank you! :)

PR #2065, opened 2023-02-03
Add support to sync preferences across browsers/devices

FireMasterK:


Atemu:
Would it perhaps be possible to not try to do e2ee in step 1?

While it'd be nice, my preferences just simply aren't critical data worth protecting to this degree.  
I'd much rather have the feature without e2ee than not at all.

t6fb3m59:
I don't quite understand why a different approach would be required for the login functionality. If I am not mistaken, the current approach is checking a bcrypt hash. I am not sure how this is exposed in the backend, but leaving an api call for password check through this method should be ok. The password would be used to generate an encryption key through PBKDF client side. I found a relevant discussion here: https://security.stackexchange.com/questions/2202/lessons-learned-and-misconceptions-regarding-encryption-and-cryptology. While it advises against this approach, it also encourages using more iterations to make it more resistent if this method is required and we should take note of this. I previously mentioned that a randomly generated encrypted encryption key would be stored server side, but could we skip this step and just use the key meant to decrypt it for the data instead? This would make things simpler server-side and we would just have an encryption key generated from the password by the client and it would handle decrypting the data without an intermediary random key stored in the db.

PR #2109, opened 2023-02-15
Option to prioritize unwatched videos when autoplaying

ai25:
The priority will be:
Next playlist video > unnwatched video from the same channel > unwatched video > first video in queue


src/components/WatchVideo.vue
```diff
@@ -556,6 +573,50 @@ export default {
         onTimeUpdate(time) {
             this.currentTime = time;
         },
+        async getWatchedRelatedVideos() {
```
vr10t, line 576: will need to remove this once #2112  gets merged

src/components/VideoPlayer.vue
```diff
@@ -38,6 +38,12 @@ export default {
                 return {};
             },
         },
+        nextVideo: {
+            type: Object,
+            default: () => {
+                return {};
+            },
+        },
```
FireMasterK, line 46: Shouldn't this be a String?

src/components/WatchVideo.vue
```diff
@@ -556,6 +573,50 @@ export default {
...
+                const watched = data.find(video => video.videoId === id);
+                if (!watched) {
+                    if (video.uploaderUrl === this.video.uploaderUrl) {
+                        this.nextVideo = video;
+                        foundAuthorMatch = true;
+                        break;
+                    } else if (!foundAuthorMatch) {
+                        this.nextVideo = video;
+                    }
+                }
```
FireMasterK, line 614: This should technically be a new preference - skipping already watched content from autoplay (maybe this could be true by default?).

src/components/VideoPlayer.vue
vr10t: I was thinking it should be a video object because we might display what's playing next when hovering on the Next button like youtube does. So for this we would need title and thumbnail as well

src/components/WatchVideo.vue
vr10t: Yeah, we could have 2 preferences for this, "Prefer unwatched videos in autoplay" and "Prefer videos from the same channel in autoplay"

src/components/VideoPlayer.vue
FireMasterK: Okay, fair enough. We could potentially move the autoplay selection logic entirely to `WatchVideo.vue` in that case!

src/components/WatchVideo.vue
```diff
@@ -206,8 +207,12 @@
                 />
                 <hr v-show="showRecs" />
                 <div v-show="showRecs">
+                    <div v-if="nextVideo" class="">
+                        <p>Next up:</p>
```
FireMasterK, line 211: This needs to be a translatable string.

vr10t:
I was confusing `autoPlayVideo` preference with `autoplay`. I added `autoplay` to the preferences page and also made it more clear the other 2 options are related to "Autoplay Next" rather than "Autoplay Video".

Bnyro:
Unfortunately, there are so many conflicts that it probably makes more sense to rewrite this from scratch than resolving the merge conflicts. Thus closing.

maboroshin:
I am waiting for this. I am waiting for a new PL.

PR #2112, opened 2023-02-16
Watch progress bar under video thumbnails

vr10t:
Closes https://github.com/TeamPiped/Piped/issues/568

### DIFF
--- a/src/components/VideoItem.vue
+++ b/src/components/VideoItem.vue
@@ -10,13 +10,24 @@
                 },
             }"
         >
-            <img
-                class="w-full aspect-video"
-                :src="item.thumbnail"
-                :alt="item.title"
-                :class="{ 'shorts-img': item.isShort }"
-                loading="lazy"
-            />
+            <div class="w-full">
+                <img
+                    class="w-full aspect-video"
+                    :src="item.thumbnail"
+                    :alt="item.title"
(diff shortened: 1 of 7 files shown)


vr10t:
Closes #568?

Bnyro (approved):

src/components/VideoItem.vue
```diff
@@ -148,6 +159,7 @@ export default {
         };
     },
     mounted() {
+        this.updateWatched([this.item]);
```
FireMasterK, line 162: Just wondering, why can't this be called in a batch? 

Like how we already do: https://github.com/search?q=repo%3ATeamPiped%2FPiped%20updateWatched&type=code

vr10t: Because you'd have to update them in in every component that uses it, as well as potential components that might use it in the future. There's nothing wrong with doing it like that but this way it's easier. Also, I think this way it doesn't make the db request until it's in view? I'm not exactly sure how Vue works to be honest

FireMasterK: But that method is already being called already in a few places, I would keep it that way to keep things fast :)

Is this line even necessary?

vr10t: All done :)

FireMasterK (approved):

PR #2118, opened 2023-02-17
Option to export/import watch history to db.

### LINKED ISSUE #1514: Import/export watched videos
### Describe the feature

When I export preferences and import them in another instance I would like to get also the video history and get the videos I've watched marked as `watched`.

To move from one instance to another without noticing anything different while using LibRedirect.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

### LINKED ISSUE #2117: Export history as Watch History
### Describe the feature

Option to export watch history to json which can then be imported and added to the database instead of as a playlist.

It would be much more convenient and it would provide a way to sync history between devices.

### Concept(s)

_No response_

### Additional context

I am aware of the privacy drawbacks of this method, that's why we should keep the playlist way as an option, as well as select which fields you want to keep when exporting as history.

### Acknowledgements

ai25:
Closes #2117
Closes #1514

### DIFF
--- a/src/components/ExportHistoryModal.vue
+++ b/src/components/ExportHistoryModal.vue
@@ -27,7 +27,7 @@
                     </label>
                 </div>
             </form>
-            <button class="btn mt-4" @click="handleExport" v-text="$t('actions.export_history')" />
+            <button class="btn mt-4" @click="handleExport" v-text="$t('actions.export')" />
         </div>
     </ModalComponent>
 </template>
--- a/src/components/ExportHistoryModal.vue
+++ b/src/components/ExportHistoryModal.vue
@@ -1,0 +1,142 @@
+<template>
+    <ModalComponent>
+        <div class="min-w-max flex flex-col">
+            <h2 class="mb-4 text-center text-xl font-bold">Export History</h2>
+            <form>
+                <div>
+                    <label class="mr-2" for="export-format">Export as:</label>
+                    <select id="export-format" v-model="exportAs" class="select">
+                        <option
+                            v-for="option in exportOptions"
+                            :key="option"
+                            :value="option"
+                            v-text="formatField(option)"
+                        />
+                    </select>
+                </div>
+                <div v-if="exportAs === 'history'">
+                    <label v-for="field in fields" :key="field" class="flex items-center gap-2">
+                        <input
+                            v-model="selectedFields"
+                            class="checkbox"
+                            type="checkbox"
+                            :value="field"
+                            :disabled="field === 'videoId'"
+                        />
+                        <span v-text="formatField(field)" />
+                    </label>
+                </div>
+            </form>
+            <button class="btn mt-4" @click="handleExport">Export</button>
+        </div>
+    </ModalComponent>
+</template>
+
+<script>
+import ModalComponent from "./ModalComponent.vue";
+
+export default {
+    components: {
+        ModalComponent,
+    },
+    data() {
+        return {
+            exportOptions: ["playlist", "history"],
+            exportAs: "playlist",
+            fields: [
+                "videoId",
+                "title",
+                "uploaderName",
+                "uploaderUrl",
+                "duration",
+                "thumbnail",
+                "watchedAt",
+                "currentTime",
+            ],
+            selectedFields: [
+                "videoId",
+                "title",
+                "uploaderName",
+                "uploaderUrl",
+                "duration",
+                "thumbnail",
+                "watchedAt",
+                "currentTime",
+            ],
+        };
+    },
+    methods: {
+        async fetchAllVideos() {
+            if (window.db) {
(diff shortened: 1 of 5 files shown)


src/components/HistoryPage.vue
```diff
@@ -5,7 +5,9 @@
         <div>
             <button class="btn" v-t="'actions.clear_history'" @click="clearHistory" />
 
-            <button class="btn mx-3" v-t="'actions.export_to_json'" @click="exportHistory" />
+            <button class="btn mx-3" v-t="'actions.export_to_json'" @click="showModal = !showModal" />
+
+            <button class="btn" v-t="'actions.import_from_json'" @click="$router.push('/history/import')" />
```
Bnyro, line 10: This should be an `a` element instead, the JS isn't needed here.

src/components/HistoryPage.vue
```diff
@@ -20,20 +22,77 @@
     </div>
 
     <br />
+    <ModalComponent v-if="showModal" @close="showModal = !showModal">
```
Bnyro, line 25: I feel like that modal should be placed in a different file to keep it easier to overview, let's see what K thinks about it.

src/components/ImportHistoryPage.vue
```diff
@@ -0,0 +1,98 @@
+<template>
```
Bnyro, line 1: I'd prefer having this as a modal instead of being its own page since the page only contains three buttons in total.

src/components/ImportHistoryPage.vue
```diff
@@ -0,0 +1,98 @@
...
+    <div class="text-center min-h-screen">
+        <h1 class="text-center my-2">Import History</h1>
+        <form>
+            <br />
+            <div>
+                <input ref="fileSelector" type="file" @change="fileChange" />
```
Bnyro, line 7: We should style that file selector similar to the one from the playlists page, see https://github.com/TeamPiped/Piped/blob/master/src/components/PlaylistsPage.vue#L13-L14.

src/components/HistoryPage.vue
FireMasterK: I agree, we should make a new component for it for future readability reasons.

src/components/ImportHistoryPage.vue
vr10t: The only reason I added it as a second page is because I felt like there was too much code that doesn't really have much to do with history, it's more about importing. Also, if we're going to support importing from other platforms in the future that will be even more code and possibly instructions, similar to the subscription import page

Bnyro: I get your point with the code structure, however that wouldn't be an issue if you just keep everything in that file and just wrap it into a `ModalComponent`. And I don't think there is that much of additional information needed as on the subscriptions page since not many other projects let you export the history currently, so instructions for others shouldn't be needed as much as on the subscriptions import page.

vr10t: You were right, it's much cleaner this way.

src/components/ExportHistoryModal.vue
```diff
@@ -0,0 +1,142 @@
...
+                    <label v-for="field in fields" :key="field" class="flex gap-2 items-center">
+                        <input
+                            class="checkbox"
+                            type="checkbox"
+                            :value="field"
+                            v-model="selectedFields"
+                            :disabled="field === 'videoId'"
+                        />
+                        <span v-text="formatField(field)" />
+                    </label>
```
FireMasterK, line 27: We likely don't want this to be user selectable, as we want to stick to a standard format between frontend implementations.

We would potentially align to a new format in the distant future to allow inter-compatibility with other frontends once a format is standardized - https://github.com/UniversalPipeWrench/unified-user-data-format/issues/1.

vr10t: I see one problem with this: What if the format we decide on is not what users want? 
For example, if we include watchedAt and currentTime by default this might be a privacy concern for some users. On the other hand, if we decide to leave them out we are throwing out features which I would assume most users want, myself included.
The only solution I can think of is for the standard format to allow these to be optional.

FireMasterK: I feel `currentTime` is always necessary since otherwise, we can't show any indicators on videos (just the watch history page).

Regarding `watchedAt`, that's a fair point, but we don't know how to sort videos then.

However, I feel this is less of an issue since watch history is stored entirely locally - so, it's unlikely any third party gets access to it in the first place.

Atemu:
What's the status on this?

UltraBlackLinux:
I'd like to bump this too. Unlike the "mark as watched" button PR, this seems to be fully ready for merge.

tunloop:
Is there any progress on this? 
Currently there is already an export watch history button, but no import?

Bnyro (approved):

PR #2119, opened 2023-02-17
Added ytify to README

n-ce:
I am not sure if piped's policy is inline with the service ytify provides but It's been a few weeks it has been using the piped API and it has provided a very stable infrastructure support for the project so I think it may have a place here in this list.

I would understand if this PR is closed, still want to thank the Piped Team & community for maintaining such an awesome API.

**Repo** : https://github.com/n-ce/ytify/
**Deploy** : https://ytify.netlify.app


FireMasterK:
Awesome, thank you for the cool project!

PR #2121, opened 2023-02-19
Show video category

vr10t:
Since the backend supports it now.

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -148,11 +148,13 @@
 
             <!-- eslint-disable-next-line vue/no-v-html -->
             <div v-show="showDesc" class="break-words" v-html="purifyHTML(video.description)" />
-            <div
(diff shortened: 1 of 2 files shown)


src/components/WatchVideo.vue
```diff
@@ -152,6 +152,7 @@
                 v-if="showDesc && sponsors && sponsors.segments"
                 v-text="`${$t('video.sponsor_segments')}: ${sponsors.segments.length}`"
             />
+            <div v-if="showDesc && video.category" v-text="`${$t('video.category')}: ${video.category}`" />
```
Bnyro, line 155: Everything that is only visible when `showDesc` is true should bee wrapped into a `template` to avoid having the same if condition multiple times in a row.

vr10t: :+1:

Bnyro (approved):

PR #2144, opened 2023-02-27
Fix LBRY playback.

FireMasterK:


src/components/VideoPlayer.vue
```diff
@@ -273,7 +273,7 @@ export default {
 
             const lbry = this.getPreferenceBoolean("disableLBRY", false)
                 ? null
-                : this.video.videoStreams.filter(stream => stream.quality === "LBRY")[0];
+                : this.video.videoStreams.filter(stream => stream.quality.indexOf("LBRY") !== -1)[0];
```
Bnyro, line 276: Just a question, why not using `stream.quality.includes("LBRY")` instead?

FireMasterK: Didn't know such a thing existed lol, I've always looked for a "contains" method, thanks!

Bnyro:
Is there anything still blocking that PR?

FireMasterK:
The problem is it doesn't actually 'fix' anything, we still get a 401 unfortunately.

I'll send odysee an email to see what happens.

Bnyro:
Are you sure? I was pretty sure it worked fine when I tested it with LibreTube :thinking:

PR #2153, opened 2023-03-01
Make preferences more visible by alternating background colors

vr10t:
This makes it more clear which checkbox belongs to which preference.


src/components/PreferencesPage.vue
```diff
@@ -602,4 +602,7 @@ export default {
 .pref {
     @apply flex justify-between items-center my-2 mx-[15vw] lt-md:mx-[2vw];
 }
+.pref:nth-child(odd) {
+    @apply bg-dark-800;
+}
```
FireMasterK, line 607: Do we need something else for light theme users?

FireMasterK (approved):

PR #2154, opened 2023-03-01
Decrease opacity of watched video thumbnails.

vr10t:
This makes it easier to see which videos are watched, especially on big screens.


FireMasterK:
I see almost no difference with this on my monitor, but if it makes a difference on big screens, that's great!

PR #2163, opened 2023-03-03
feat: reactive title bar based on user's theme

### LINKED ISSUE #2139: Safari PWA title bar should be black
### Describe the feature

Change the title bar to black instead of the current red color that is very distracting

1E99E588-62D1-4D6F-9D3C-CE70294B225B

Using the PWA on iOS diverts my eyes to the top bright red color and it’s very distracting

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

rramiachraf:
Fixes #2139

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -34,10 +34,12 @@ export default {
             if (themePref == "auto") this.theme = darkModePreference.matches ? "dark" : "light";
             else this.theme = themePref;
 
-            // Change title bar color to white
+            // Change title bar color based on user's theme
+            const themeColor = document.querySelector("meta[name='theme-color']");
             if (this.theme === "light") {
-                const themeColor = document.querySelector("meta[name='theme-color']");
                 themeColor.setAttribute("content", "#FFF");
+            } else {
+                themeColor.setAttribute("content", "#0F0F0F");
             }
 
             // Used for the scrollbar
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -34,6 +34,12 @@ export default {
             if (themePref == "auto") this.theme = darkModePreference.matches ? "dark" : "light";
             else this.theme = themePref;
 
+            // Change title bar color to white
+            if (this.theme === "light") {
+                const themeColor = document.querySelector("meta[name='theme-color']");
+                themeColor.setAttribute("content", "#FFF");
+            }
```
FireMasterK, line 41: Don't we need to set it back to dark if the theme changes back to dark? (eg: when using auto mode)

rramiachraf: The `theme-color` meta tag by default uses the dark theme, therefore if dark theme is set the condition would be false and the theme color won't be set which means a fallback to the default value.

rramiachraf: Another approach would be injecting the meta tag conditionally based on user's theme.

FireMasterK: What if this scenario:

- The user has auto theme mode and is currently using dark mode
- The user changes their theme from their system preferences to light, and then to dark (without reloading the page)

Won't it remain light in such a scenario?

rramiachraf: You're absolutely right, forgot about this scenario.
Next commit will fix this.

FireMasterK (approved):
Thank you!

Issue #2159, opened 2023-03-02
Add placeholders/loading indicators


vr10t:
### Describe the feature

Provide visual feedback for loading components.

### Why would this be useful to add?

Loading spinners are useful for feedback and placeholders would prevent layout shifts.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

PR #2173, opened 2023-03-06
fix: make footer stick to the bottom

rramiachraf:
The current behavior is not visually appealing, and is more noticeable on pages with small amount of content like the register and login pages.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,14 +1,13 @@
 <template>
     <div class="flex flex-col w-full min-h-screen px-1vw py-5 reset" :class="[theme]">
-        <NavBar />
-
-        <main class="flex-1">
+        <div class="flex-1">
+            <NavBar />
             <router-view v-slot="{ Component }">
                 <keep-alive :max="5">
                     <component :is="Component" :key="$route.fullPath" />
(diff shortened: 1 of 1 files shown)


rramiachraf:
After this commit I've noticed a problem with the search suggestions menu, it started getting on top of the search input making it harder for users to see which letters are being typed.
A quick fix by using `top: 4rem` would solve this but it seems like a hacky way of achieving things.

Bnyro:
I can reproduce the search suggestions issue, couldn't we just give the `main` a `min-height: 80vh` or something? Sure, that wouldn't be as elegant as wrapping it into a flex-container, but it should at least get rid of the issue with the search suggestions.

rramiachraf:
Adding the navbar component inside `main` seems to get rid of the problem.

Bnyro:
Honestly I'm not sure if it looks better like that as we currently don't have any loading indicator so while loading, there's just a big white/black gap in-between the footer and the top bar.
I think this would be an awesome improvement in combination with #2159.

Bnyro (approved):

Bnyro:
Should be merged after #2192 imo

PR #2192, opened 2023-03-12
Loading indicators

### LINKED ISSUE #2159: Add placeholders/loading indicators

Bnyro:
closes #2159

### DIFF
--- a/src/components/FeedPage.vue
+++ b/src/components/FeedPage.vue
@@ -24,32 +24,29 @@
 
     <hr />
 
-    <div v-if="videos.length != 0" class="video-grid">
+    <LoadingIndicatorPage :show-content="videosStore != null" class="video-grid">
         <template v-for="video in videos" :key="video.url">
             <VideoItem v-if="shouldShowVideo(video)" :is-feed="true" :item="video" />
         </template>
-    </div>
-    <div v-else>
-        <LoadingIndicator />
-    </div>
+    </LoadingIndicatorPage>
 </template>
 
 <script>
 import VideoItem from "./VideoItem.vue";
 import SortingSelector from "./SortingSelector.vue";
-import LoadingIndicator from "./LoadingIndicator.vue";
+import LoadingIndicatorPage from "./LoadingIndicatorPage.vue";
 
 export default {
     components: {
         VideoItem,
         SortingSelector,
-        LoadingIndicator,
+        LoadingIndicatorPage,
     },
     data() {
         return {
             currentVideoCount: 0,
             videoStep: 100,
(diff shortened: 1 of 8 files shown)


rramiachraf:
Great work!

Since there's code splitting and routes are dynamically imported it might be a good idea to also add a loading indicator for them too.
A common scenario is when the application is not cached on the user's device and they try to navigate to the login route, when the link gets clicked it looks like nothing happened and suddenly the page pops out.
One way to approach this is by adding navigation guards to the router, `beforeEach` and `afterEach` would solve this perfectly.

Bnyro:
That's a good idea, feel free to merge my changes into a local branch of yours, do the changes and create a new PR that superceeds mine since it seems like you already have some ideas how to get it working properly! :)

rramiachraf:
I'll wait until this and #2173 get merged, I messed up my workflow by working directly on the `master` branch on my fork resulting all the upcoming PRs including unmerged changes from `master`.
Here's one I've created and immediately deleted it :p https://github.com/TeamPiped/Piped/pull/2181

src/components/FeedPage.vue
```diff
@@ -24,21 +24,26 @@
 
     <hr />
 
-    <div class="video-grid">
+    <div v-if="videos.length != 0" class="video-grid">
```
ChunkyProgrammer, line 27: Maybe a different condition should be checked? Ex: what if a user isn't subscribed to any channel (maybe they forgot to login). It will just seem like the page is loading forever.

Bnyro: True, good point!

Bnyro:
Haha, I feel you ;)
Though I saw you added a dependency for the loading indicator, are you sure that'd really be a good idea? I feel like a progress indicator is something we can deal with without any third party dependency since we have more control over it like so.

FireMasterK (commented):
I thought of a different approach than `v-if`: Why not add a slot for the `LoadingIndicator` component, and maybe add the condition as a prop?

(We'd still need a v-if inside the `LoadingIndicator` component tho)

Bnyro:
What would be the advantage of it? I feel like it wouldn't make much of a difference at all, so it's just for better code readability mostly?

FireMasterK:
> What would be the advantage of it? I feel like it wouldn't make much of a difference at all, so it's just for better code readability mostly?

Yes, it is just slightly better to read.

PR #2199, opened 2023-03-14
Frame-by-frame navigation

vr10t:
Navigate one frame at a time using `,` and `.`

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -100,7 +100,7 @@ export default {
         this.hotkeysPromise.then(() => {
             var self = this;
             this.$hotkeys(
-                "f,m,j,k,l,c,space,up,down,left,right,0,1,2,3,4,5,6,7,8,9,shift+n,shift+,,shift+.,return",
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
LGTM!

Bnyro:
I had to revert the changes due to #2206, I guess that using multiples `,` in line 103 causes a conflict unfortunately.

vr10t:
Hmm, that's odd.. [This](https://github.com/jaywcjlove/hotkeys/blob/862a224672514e392e9ce55a9058429f67122ad5/src/utils.js#L20) is the function that splits the keys in hotkeys-js and from testing, it should parse the string just fine.
Here is the output:
`0:f` `1:m` `2:j` `3:k` `4:l` `5:c` `6:space` `7:up` `8:down` `9:left` `10:right` `11:0` `12:1` `13:2` `14:3` `15:4` `16:5` `17:6` `18:7` `19:8` `20:9` `21:shift+n` `22:shift+,` `23:shift+.` `24:return` `25:.` `26:,`

vr10t:
No problem! We should probably add a comment though, to make sure the keys get split properly if we change them or add more in the future.

PR #2204, opened 2023-03-15
Tweak channel page layout

rramiachraf:
Before:
before

After:
after

### DIFF
--- a/src/components/ChannelPage.vue
+++ b/src/components/ChannelPage.vue
@@ -43,11 +43,11 @@
         </div>
 
         <!-- eslint-disable-next-line vue/no-v-html -->
-        <div class="whitespace-pre-wrap py-2">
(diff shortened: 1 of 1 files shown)


Bnyro:
I have three minor suggestions for improvements:
* Keep the horizontal margin symmetric and the same for everything (channel icon, description, video filters) as they currently aren't in a vertical light on the left
* Please make the "Show more" and "Show less" strings translatable or use any already existing string for it (see `src/locales/en.json`).
* Center align the "Show more"/"Show less" button, slightly increase its top padding and decrease its opacity as I think it should be less striking/eye-catching than the actual description

Apart from that I really like the changes :)

rramiachraf:
- [x] Make "Show more" and "Show less" strings translatable.
- [x] Decrease their opacity.

@Bnyro Could you please elaborate further on the first suggestion and the first part from the third one?

Issue #1056, opened 2022-05-16
[ Feature Request ] Alt+P Shortcut for Picture-in-Picture


8227846265:
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.kavin.rocks/) or is API related.

### Describe the bug

Hello, I would like to suggest adding a keyboard shortcut (e.g. Alt + P) to enable/disable the Picture-in-Picture mode.

### To Reproduce

To Enable
* Watch a video
* Press Alt+P
* PiP will be enabled

To Disable
* Go back to the tab of the video playing PiP video
* Press Alt+P
* PiP will be disabled

While PiP is enabled, and you want to PiP another video
* Watch a video
* Press Alt+P
* PiP will be enabled
* Switch to another tab with a video
* Press Alt+P
* The current PiP window will be replaced with the new one.

### Expected behavior

As seen above.

### Logs/Errors

None

### Browser, and OS with Version.

All

PR #2218, opened 2023-03-16
Fix: Add a hotkey to toggle Picture-In-Picture

EscherMoore:
This PR closes issue #1056. It uses the Picture-in-Picture web API, which is not yet supported on some browsers. For a compatibility list of the Picture-in-Picture web API, please refer to the following link: [https://developer.mozilla.org/en-US/docs/Web/API/Document/pictureInPictureElement#browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Document/pictureInPictureElement#browser_compatibility).

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -100,7 +100,7 @@ export default {
         this.hotkeysPromise.then(() => {
             var self = this;
             this.$hotkeys(
-                "f,m,j,k,l,c,space,up,down,left,right,0,1,2,3,4,5,6,7,8,9,shift+n,shift+,,shift+.,return,.,,",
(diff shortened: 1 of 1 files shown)


FireMasterK:
If it is unsupported, you should use `requestPictureInPicture()`

For example like `document.querySelector("video").requestPictureInPicture()`

See related shaka code: https://github.com/shaka-project/shaka-player/blob/62eea576601dcaef4839bee0a1fab6d748904d6e/ui/pip_button.js#L128-L140

EscherMoore:
The code in this PR already makes use of the same `requestPictureInPicture()` method as in the provided Shaka source code. The `requestPictureInPicture()` method used in the Shaka code is, sadly, also a part of the standard Picture-in-Picture Web API and is not a custom Shaka method that circumvents the standard Picture-in-Picture Web API and overcomes its limitations regarding browser support. Please correct me if I am misunderstanding. Thanks!

PR #2258, opened 2023-03-26
Avoid unnecessary precaching

### LINKED ISSUE #1600: All TimeAgo files cached by ServiceWorker
All TimeAgo language files are cached/stored by the service worker, we should cache them on demand instead to save bandwidth (Close to 1MB?!).

snematoda:
- Closes #1600
- Disabled preaching of all favicons https://developer.chrome.com/docs/workbox/precaching-dos-and-donts/#dont-precache-responsive-images-or-favicons

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -21,7 +21,7 @@ export default defineConfig({
         VitePWA({
             registerType: "autoUpdate",
             workbox: {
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Lgtm, thanks :+1:

PR #2310, opened 2023-04-18
Replaced "delete-search" icon as (⨉) instead of (x)

SecularSteve:
It looks so much better...

Screenshot 2023-04-18 at 23-10-10 Piped


FireMasterK (approved):

Bnyro:
Imo, the padding should have been increased and the font size decreased too because the X now fills pretty much the whole circle.

PR #2353, opened 2023-05-01
Implement support for feed pagination

### LINKED ISSUE #1046: Pagination for subscription feeds
Currently, when a user has a large number of subscriptions, the feed response takes multiple seconds to generate

What can be done is limiting each page to 100 results, and adding an API for nextpage calls.

The problem is that this would require breaking API changes. (It would affect other projects like LibreTube, Yattee, Piped-Material, etc)

FireMasterK:
Closes #1046
See https://github.com/TeamPiped/Piped-Backend/pull/583


FireMasterK:
Converted to draft since we now require some changes since #2372

PR #2355, opened 2023-05-01
Fix seekbar preview aspect ratio and cropping

GlenLowland:
This PR fixes two issue observed with the seekbar preview:
- sometimes images got cropped incorrectly at the end of the video due to image having less frames that server tells us - so calculate it in place;
- set canvas width and height based on actual frame data so it follows the video aspect ratio rather than hard-coded value.


FireMasterK:
Thank you!

PR #2363, opened 2023-05-05
Migrate project to pnpm.

### LINKED ISSUE #2362: Migrate the project to pnpm from yarn
We should do this, so it would become faster to install dependencies and pnpm is space efficient when developing multiple projects.

See https://pnpm.io/benchmarks

FireMasterK:
Closes #2362


Bnyro:
So if I get it correctly, pnpm installs dependencies system wide and not only per project?

FireMasterK:
It actually installs them in its own directory in the home folder and then uses symlinks to create a `node_modules` folder as far as I understand it. That way, you don't need to store the same file multiple times. In Piped it doesn't matter much as our application is quite small as we have only one `package.json` project, but it has great potential in monorepos!

PR #2372, opened 2023-05-07
Subscription groups

### LINKED ISSUE #490: Request: categories
Using an old computer to use youtube has become very slow, Piped is extraordinary in comparison.

I have two main issues:
- I couldn't import the youtube subscriptions from my account. First because using Google Takout I only get an archive with an html file that says "no data found". (I tried a different youtube account I have, that too "no data found", maybe the system is currently broken) Secondly, even if it worked it's now a .cvs format and not .json

- This is the actual request. Due to a lot of subscriptions, I was using browser extension to add "categories". So instead of browsing the general feed page I could select a category I made, and I could so filter the subscriptions under each category. 

It would be quite useful and I suppose it can only be stored locally. But that would be already fine.

Bnyro:
closes #490

(Note that they are NOT synced with your account and can not yet become backup'ed or restored!)


src/components/FeedPage.vue
```diff
@@ -18,14 +18,27 @@
         <option v-for="filter in availableFilters" :key="filter" :value="filter" v-t="`video.${filter}`" />
     </select>
 
+    <label for="group-selector" class="ml-10 mr-2">
+        <strong v-text="`${$t('titles.channel_groups')}:`" />
+    </label>
+    <select id="group-seletor" v-model="selectedGroupName" default="" class="select w-auto">
```
FireMasterK, line 24:
```suggestion
    <select id="group-selector" v-model="selectedGroupName" default="" class="select w-auto">
```

Bnyro:
Looks like I also forgot to run `git add` before :upside_down_face:

Bnyro:
If you have a better idea than the current `DefaultValueCheckbox`, that would be great.
Using `:value` doesn't work, I tested it.

src/components/SubscriptionsPage.vue
```diff
@@ -36,20 +56,70 @@
...
+        if (!window.db) return;
+        const cursor = this.getChannelGroupsCursor();
+        cursor.onsuccess = e => {
+            const cursor = e.target.result;
+            if (cursor) {
+                const group = cursor.value;
+                this.channelGroups = this.channelGroups.concat({
+                    groupName: group.groupName,
+                    channels: JSON.parse(group.channels),
+                });
```
FireMasterK, line 120: Why not just `this.channelGroups.push`? I noticed that if we have more than 1 channel group, only one seems to show, despite being in the database 🤔

vr10t:
> If you have a better idea than the current `DefaultValueCheckbox`, that would be great.
> Using `:value` doesn't work, I tested it.

You can use the `:checked` property

src/components/SubscriptionsPage.vue
Bnyro: There were some `cursor.continue` calls missing which caused that. 
I didn't use push because I was still used to Svelte, which requires you to assign a new value in order to update. (push wouldn't work in Svelte).

Bnyro:
> You can use the `:checked` property
>
Oh well, thank you!
If I knew that earlier or the docs mentioned it that could have saved me a lot of time :upside_down_face:

vr10t:
> > You can use the `:checked` property
> >
> Oh well, thank you!
> If I knew that earlier or the docs mentioned it that could have saved me a lot of time :upside_down_face: 
> 

No worries, I didn't know either. I knew React has `defaultChecked` so I just searched for 'Vue defaultChecked'

FireMasterK (approved):

PR #2399, opened 2023-05-15
Fix the play button not working consistently on mobile devices

### LINKED ISSUE #2395: Play Button Broken on Mobile
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.video/) or is API related.

### Describe the bug

I've always been unable to watch videos using Piped on my Android phone (using Bromite). I figured out that the play button is "broken" – I need to tap slightly to the right/top of it in order to start any given video.

### To Reproduce

1. Open any video via Piped on mobile
2. The video does not play automatically, even though the default setting is to do so (AFAIK, the default setting should play the video upon opening a Piped video link)
6. Chromium browsers (Bromite, DuckDuckGo): Play button only responds when tapping to the right/above it. Fennec: Play Button doesn't respond at all.

### Expected behavior

Video is played automatically, or at least upon tapping on the play button in the center of the frame.

### Logs/Errors

N/A

Chromium 113.0.5672.61 on Android 13
DuckDuckGo 5.156.1 sc
Fennec F-Droid 112.2.0 (Build #1122020), 7d1493c332+

### Additional context

Didn't find a similar bug after a quick search, but I can't imagine no one reported it if it really is as prevalent as I'm assuming it to be. At the same time, I'm not aware of doing something incorrectly myself.

FireMasterK:
Closes #2395

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -854,5 +854,6 @@ export default {
     bottom: 0;
     margin-bottom: 4.5%;
     border-radius: 0.3rem;
+    display: none;
 }
 </style>


IronMaltese:
Thank you!!!

Issue #2423, opened 2023-05-19
Add default SPEED and default CAPTION ON


p0mad:
### Describe the feature

It is likely happen that when i want to watch videos, always change the speed to 1.75 or 2x speed and also for non english speakers it is good to add always caption ( english ) ON option for the setting like whats available for quality of the videos

### Why would this be useful to add?

As i said most people tend to listen or watch in 2X speed for all the videos

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

PR #2427, opened 2023-05-20
Add option to display automatically the subtitles

### LINKED ISSUE #2295: setting to show captions by default
### Describe the feature

I have to set the caption manually to be shown every time I load a video as it is Off be default.
can this be change to user set in setting?
for example "show captions always for (non-generated) English" or any language like that that video actually has captions for it?

it removed the need for user to manually set it for every video he watches.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

judemont:
Hello,
I have added an option in the preferences to automatically display the subtitles.
### Issues :
closes #2295
touches #2423

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -587,6 +587,11 @@ export default {
                 const rate = this.getPreferenceNumber("rate", 1);
                 videoEl.playbackRate = rate;
                 videoEl.defaultPlaybackRate = rate;
+
+                const autoDisplayCaptions = this.getPreferenceBoolean("autoDisplayCaptions", false);
+                if (autoDisplayCaptions) {
+                    this.$player.setTextTrackVisibility(!this.$player.isTextTrackVisible());
+                }
             });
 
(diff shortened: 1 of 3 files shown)


src/components/VideoPlayer.vue
```diff
@@ -587,6 +587,11 @@ export default {
                 const rate = this.getPreferenceNumber("rate", 1);
                 videoEl.playbackRate = rate;
                 videoEl.defaultPlaybackRate = rate;
+
+                const autoDisplayCaptions = this.getPreferenceBoolean("autoDisplayCaptions", false);
+                if (autoDisplayCaptions) {
+                    this.$player.setTextTrackVisibility(!this.$player.isTextTrackVisible());
```
Bnyro, line 593: Shouldn't this just be `setTextTrackVisibility(true)`?

judemont: Yes that's right, sorry for the confusion.

Bnyro (approved):

FireMasterK (approved):
Thank you!

PR #2455, opened 2023-05-25
fix safari "SyntaxError: Invalid regular expression: invalid group sp…

robertkleinschuster:
…ecifier name" caused by #2002 from Tyrritt/comment-timestamps-seek

### DIFF
--- a/src/components/WatchVideo.vue
+++ b/src/components/WatchVideo.vue
@@ -236,6 +236,7 @@ import PlaylistVideos from "./PlaylistVideos.vue";
 import WatchOnYouTubeButton from "./WatchOnYouTubeButton.vue";
 import LoadingIndicatorPage from "./LoadingIndicatorPage.vue";
 import ToastComponent from "./ToastComponent.vue";
+import { parseTimeParam } from "@/utils/Misc";
 
 export default {
     name: "App",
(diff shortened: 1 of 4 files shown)


src/components/WatchVideo.vue
```diff
@@ -540,16 +540,21 @@ export default {
...
+            if (this.handleTimestampLinks(target)) {
+                event.preventDefault();
+            }
+        },
+        handleTimestampLinks(target) {
+            const url = new URL(target.getAttribute("href"), document.baseURI);
+            if (!url.searchParams.has("t")) return false;
+            const time = parseInt(url.searchParams.get("t"));
+            url.searchParams.delete("t");
+            if (document.location.href !== url.href) return false;
```
FireMasterK, line 555: The current video could have a lot of parameters, causing this to fail at times! Please instead check the current video id along with the `v` parameter from the URL.

FireMasterK (changes requested):

src/components/WatchVideo.vue
```diff
@@ -540,16 +540,21 @@ export default {
...
+            if (!target.getAttribute("href")) return;
+            if (this.handleTimestampLinks(target)) {
+                event.preventDefault();
+            }
+        },
+        handleTimestampLinks(target) {
+            const url = new URL(target.getAttribute("href"), document.baseURI);
+            if (!url.searchParams.has("t")) return false;
+            const time = parseInt(url.searchParams.get("t"));
+            url.searchParams.delete("t");
```
FireMasterK, line 554: There's no need to remove the parameter from the search params here.

src/components/WatchVideo.vue
```diff
@@ -540,16 +540,21 @@ export default {
...
+            const target = event.target;
+            if (!target.getAttribute("href")) return;
+            if (this.handleTimestampLinks(target)) {
+                event.preventDefault();
+            }
+        },
+        handleTimestampLinks(target) {
+            const url = new URL(target.getAttribute("href"), document.baseURI);
+            if (!url.searchParams.has("t")) return false;
+            const time = parseInt(url.searchParams.get("t"));
```
FireMasterK, line 553: It may not be safe to assume `t` is a number. I remember there was a function somewhere to parse this.

FireMasterK (approved):
Looks great, thank you!

PR #2457, opened 2023-05-26
Improved appearance of nav bar on mobile

robertkleinschuster:
Fullwidth clickable, icons, seperators:

Bildschirm­foto 2023-05-26 um 10 48 29
Bildschirm­foto 2023-05-26 um 10 49 48

Search bar equal width to video, focus state outline instead of border to prevent shift of placeholder text

Bildschirm­foto 2023-05-26 um 10 52 07

### DIFF
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -195,7 +195,7 @@ export default {
     @apply absolute right-3 cursor-pointer rounded-full bg-[#ccc] w-4 h-4 text-center text-black opacity-50 hover:(opacity-70) text-size-[13px];
     line-height: 1.05;
 }
-.mobile-nav svg {
-    width: 16px;
+.mobile-nav div {
+    @apply mx-1;
 }
 </style>
--- a/src/components/NavBar.vue
+++ b/src/components/NavBar.vue
@@ -59,35 +59,41 @@
         </ul>
     </nav>
     <!-- navigation bar for mobile devices -->
-    <ul
+    <div
         v-if="showTopNav"
-        class="flex flex-col justify-center items-end mb-4 children:(my-0.5 mr-5)"
-        @click="showTopNav = false"
+        class="mobile-nav flex flex-col mb-4 children:(p-1 w-full border-b border-dark-100 flex items-center gap-1)"
     >
-        <li v-if="shouldShowTrending">
-            <router-link v-t="'titles.trending'" to="/trending" />
-        </li>
-        <li>
-            <router-link v-t="'titles.preferences'" to="/preferences" />
-        </li>
-        <li v-if="shouldShowLogin">
-            <router-link v-t="'titles.login'" to="/login" />
-        </li>
-        <li v-if="shouldShowLogin">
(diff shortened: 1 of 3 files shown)


Bnyro:
The title should rather be `Improved appearance of nav bar on mobile` I guess. Also, I personally don't see the point of adding icons to the navigation bar, I'm not aware of any other well known project that does that and I feel like it doesn't really fit well there. Just my POV though.

Bnyro:
And please force push / squash commits instead of using the same commit message three times.
That way we now would need to squash merge your PR so that the commit history is clean. 
(Not much of an issue here though, but it can be one of multiple commits do essentially different things).

robertkleinschuster:
> And please force push / squash commits instead of using the same commit message three times.
> That way we now would need to squash merge your PR so that the commit history is clean.
> (Not much of an issue here though, but it can be one of multiple commits do essentially different things).

I will squash them, sorry it‘s a habit of mine because of convention at my day job 😆

robertkleinschuster:
> The title should rather be `Improved appearance of nav bar on mobile` I guess. Also, I personally don't see the point of adding icons to the navigation bar, I'm not aware of any other well known project that does that and I feel like it doesn't really fit well there. Just my POV though.



> The title should rather be `Improved appearance of nav bar on mobile` I guess. Also, I personally don't see the point of adding icons to the navigation bar, I'm not aware of any other well known project that does that and I feel like it doesn't really fit well there. Just my POV though.

I thought the icons looked nice and i got the idea from the YouTube menu. 

But i can remove them for the time being and we can discuss this separately because the main issue I wanted to fix, is that the links are easy to miss on a touchscreen because they don’t extend over the whole width.

FireMasterK:
> The title should rather be `Improved appearance of nav bar on mobile` I guess. Also, I personally don't see the point of adding icons to the navigation bar, I'm not aware of any other well known project that does that and I feel like it doesn't really fit well there. Just my POV though.

I on the contrary feel they're a good thing and help make the site more intuitive to use. :thinking: I just hope it doesn't make the UI feel too crowded.

I've seen a few sites do this, even the official YouTube mobile website (when you click on the account icon, although it technically isn't a navbar).

src/components/NavBar.vue
```diff
@@ -59,35 +59,41 @@
...
-        </li>
-        <li>
-            <router-link v-t="'titles.playlists'" to="/playlists" />
-        </li>
-        <li v-if="!shouldShowTrending">
-            <router-link v-t="'titles.feed'" to="/feed" />
-        </li>
-    </ul>
+        <router-link v-if="shouldShowTrending" to="/trending">
+            <font-awesome-icon :icon="['fas', 'fire']" />
```
FireMasterK, line 67: Can we use unocss for the icons instead? I've been wanting to remove the font awesome dependency for a while.

For example, can use a `div` element with the class `i-fa6-solid:fire` in this case.

FireMasterK (approved):
LGTM!

Bnyro (approved):

Bnyro:
Thank you!

PR #2541, opened 2023-06-13
Option to disable search suggestions

### LINKED ISSUE #2533: Disabling search suggestions
### Describe the feature

Option in the settings to disable search box input being sent to the server as the search query is still being typed out, only query the server when the user hits Enter.

Search suggestions essentially turn a search box into a keylogger. Everything the user types in is immediately sent to the Piped API and then to Google. Some users (including me) are not happy with this behavior, so I think there should be an option in the settings to disable it. I believe that a search query should only be submitted when the user explicitly intends to execute the search function by pressing 'Enter'. Also, I believe it should be opt-in (so turned off by default) to save bandwidth and prevent Piped Backend Servers from being rate limited.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #2533

### DIFF
--- a/src/components/PreferencesPage.vue
+++ b/src/components/PreferencesPage.vue
@@ -136,6 +136,16 @@
             @change="onChange($event)"
         />
     </label>
+    <label class="pref" for="chkShowSearchSuggestions">
+        <strong v-t="'actions.show_search_suggestions'" />
+        <input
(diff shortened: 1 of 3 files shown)


FireMasterK (approved):
Thank you!

PR #2553, opened 2023-06-15
Playlists without an account

### LINKED ISSUE #2191: Playlists without an account
### Describe the feature

I currently use Piped as a PWA without an account. Being able to have my subscriptions and preferences available offline as well as the ability to restore them from a file when needed is extremely helpful.

It would be great if this functionality extended to playlists, so that the gap between people who use accounts on Piped and those who prefer to use it without one becomes smaller.

The feature should also allow you to backup/restore playlists, just like preferences and subscriptions.

It would allow people who would rather not create a Piped account to use more of Piped's features that people with accounts can use.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #2191

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -316,10 +316,10 @@ const mixin = {
         },
         // needs to handle both, streamInfo items and streams items
         createLocalPlaylistVideo(videoId, videoInfo) {
-            if (videoInfo === undefined || videoId === null) return;
+            if (videoInfo === undefined || videoId === null || videoInfo?.error) return;
 
-            var tx = window.db.transaction("playlistVideos", "readwrite");
-            var store = tx.objectStore("playlistVideos");
+            var tx = window.db.transaction("playlist_videos", "readwrite");
+            var store = tx.objectStore("playlist_videos");
             const video = {
                 videoId: videoId,
                 title: videoInfo.title,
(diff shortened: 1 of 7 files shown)


src/main.js
```diff
@@ -292,6 +295,252 @@ const mixin = {
...
+            var store = tx.objectStore("playlist_videos");
+            const req = store.openCursor(videoId);
+            let video = null;
+            req.onsuccess = e => {
+                video = e.target.result.value;
+            };
+            while (video == null) {
+                await this.delay(10);
+            }
+            return video;
```
FireMasterK, line 349: I really don't like this Busy Waiting approach. Why is this necessary?

Bnyro: Because IndexedDB doesn't support `Promise` and the async API. If we didn't do it that way, we would need to do a lot more changes on each playlist component to make it work with adding items one by one too. If you have a better idea to wait until it's done, instead of using short delays, that might be worth it, but I don't think we can avoid it without causing new issues.

Jeidnx: Since the function is async anyways you could use something like this:
```js
async getLocalPlaylistVideo(videoId) {
  return new Promise((resolve, reject) => {
    [...]
    req.onsuccess = e => {
      resolve(e.target.result.value);
    }
  })
}
```

Bnyro: Works like a charm, thanks!

PR #2579, opened 2023-06-20
Fix local playlist import

### LINKED ISSUE #2578: Importing Playlist while not logged in doesn't work
### Official Instance

- [X] The bug is reproducible on the [official hosted instance](http://piped.video/), or is API-related.

### Describe the bug

It seems that you can now create and export playlists while not logged in to an account, but even though the JSON export includes the links to the videos in question, importing them seems to import a playlist (though the page has to be refreshed after you import it for it to show up) but it contains zero videos.

### To Reproduce

While not logged in:

1. Create a playlist
2. Add a couple of videos
3. Export said playlist
4. Look at JSON file and note the links are there
5. Try to import the playlist
6. Note the playlist shows up, click on it
7. Note there are zero videos in it

### Expected behavior

The videos in the playlist should be properly imported.

### Logs/Errors

_No response_

Brave on Linux

### Additional context

_No response_

snematoda:
Closes #2578

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -484,13 +484,13 @@ const mixin = {
             if (!this.authenticated) {
                 const playlist = await this.getLocalPlaylist(playlistId);
                 const currentVideoIds = JSON.parse(playlist.videoIds);
-                if (currentVideoIds.length == 0) playlist.thumbnail = videoInfos[0].thumbnail;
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Thank you!

PR #2595, opened 2023-06-25
improve music artists

snematoda:
- Added `music_artists` to search filter (piped has channel tab support now)
- fixed height for ChannelItems


FireMasterK (approved):
Awesome, thank you!

Issue #2498, opened 2023-06-04
Seekbar improvements - time and focus


raisedintheusa:
### Describe the feature

Suggestion to add a time marker along with the preview when hovering over the seekbar, instead of having to click and drag to find the desired time in the video. Also, the seekbar preview currently covers up the time when hovering near the beginning of the video, again making it difficult to find a specific time.

Also, when clicking on the seekbar, pressing the spacebar to play also scrolls the page down and other shortcuts like JKL do not work since the focus has been shifted away from the video player. Seems the focus should remain on the video player when clicking the seekbar.

### Why would this be useful to add?

More quickly and easily skip to the desired time in the video.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

yonikek:
Seems the feature has been added but it's bugged on light mode.
![Screenshot from 2023-07-14 00-31-08](https://github.com/TeamPiped/Piped/assets/92832837/2d55efe2-eaa1-4653-8110-9cc7db191740)

PR #2602, opened 2023-06-27
Show timestamp when hovering seekbar

Bnyro:
see #2498

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -2,11 +2,18 @@
     <div
         ref="container"
         data-shaka-player-container
-        class="w-full max-h-screen flex justify-center"
+        class="w-full max-h-screen flex justify-center relative"
         :class="{ 'player-container': !isEmbed }"
     >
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer.vue
```diff
@@ -2,11 +2,18 @@
...
-        class="w-full max-h-screen flex justify-center"
+        class="w-full max-h-screen flex justify-center relative"
         :class="{ 'player-container': !isEmbed }"
     >
         <video ref="videoEl" class="w-full" data-shaka-player :autoplay="shouldAutoPlay" :loop="selectedAutoLoop" />
-        <canvas id="preview" />
+        <span
+            ref="previewContainer"
+            id="preview-container"
+            class="hidden flex-col absolute bottom-0 z-[2000] mb-[4.5%] items-center"
```
FireMasterK, line 12: I find that `mb-[3.5%]` looks better on my screen. I guess we need to test this on different view ports :thinking:

PR #2672, opened 2023-07-16
Apply rounded border to channels in search results

### LINKED ISSUE #2281: circular channel thumbnail in search results
### Describe the feature

current channel card in search results looks bit unappealing. also it have more height and width then the video results.

REDESIGN channel card.
adjust height & width of channel card with rounded thumbnail.

for Reference/inspiration:

also in addition **subscribers count/subscribe button** can also be optional thing to add, if others users/ devs are agree.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

iatenine:
Built per request in linked issue
Closes #2281 

At the least it provides a way to easily distinguish channels from videos in search results outside of height differences

### DIFF
--- a/src/components/ChannelItem.vue
+++ b/src/components/ChannelItem.vue
@@ -1,7 +1,7 @@
 <template>
-    <div>
+    <div class="flex flex-col flex-justify-between">
         <router-link :to="props.item.url">
-            <div class="flex justify-center h-30">
+            <div class="flex justify-center h-10-rem mb-4">
                 <img class="aspect-square rounded-full" :src="props.item.thumbnail" loading="lazy" />
             </div>
             <p>
--- a/src/components/ChannelItem.vue
+++ b/src/components/ChannelItem.vue
@@ -1,7 +1,7 @@
 <template>
     <div class="flex flex-col flex-justify-between">
         <router-link :to="props.item.url">
-            <div class="flex justify-center h-10-rem mb-4">
+            <div class="flex justify-center h-[10rem] mb-4">
                 <img class="aspect-square rounded-full" :src="props.item.thumbnail" loading="lazy" />
             </div>
             <p>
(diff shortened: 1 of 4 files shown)


Bnyro:
Instead of normal CSS, you should use the unocss class `rounded-full`.

Bnyro:
Or `rounded-[100%]`.

iatenine:
> Instead of normal CSS, you should use the unocss class `rounded-full`.

Done!

New to the repo (and Vue), wasn't sure which CSS framework was being used behind the scenes or if such a utility class existed

Bnyro:
I'm fine with the changes in general, however we should limit the width of the channel avatar and limit its width.
For reference what it currently looks like:
20230716_18h18m22s_grim

iatenine:
> I'm fine with the changes in general, however we should limit the width of the channel avatar and limit its width. For reference what it currently looks like: 20230716_18h18m22s_grim

I think h-30 on the container works pretty well

Screenshot from 2023-07-16 13-28-15
channel-listing-integrated

Bnyro:
Agreed!
Adding `my-5` or `my-10` should make it look less squeezed, feel free to check what looks best.

Bnyro:
You should use `h-[10rem]` instead of the custom rule.

iatenine:
Tried a more involved approach this time but think you'll like it

Channels, videos and playlists have been placed in flexbox containers to simplify the process of spacing between items in each result
Screenshot from 2023-07-16 15-02-27

Having found 10rem to be the sweetspot for thumbnail heights on different screens, I added a rule to the Uno classes per their documentations so we can no use "h-<num>-rem" classes to specify height in rem
Screenshot from 2023-07-16 15-05-12

I'll admit to the new channels search feeling a bit "empty" by comparison but find the smaller icons go a long way in reducing pixelation. Perhaps the video grid could be modified to include more than 5 entries per row or a visible outline border attached  but it seems like a separate issue
Screenshot from 2023-07-16 15-01-57

iatenine:
> You should use `h-[10rem]` instead of the custom rule.

Oh, that's how you do that!

Bnyro (approved):
Lgtm, thank you!

PR #2679, opened 2023-07-17
fix: hide footer and navbar when playing embedded video

### LINKED ISSUE #2438: Problems when trying to embed a video on a site
### Official Instance

- [X] The bug is reproducable on the [official hosted instance](http://piped.video/) or is API related.

### Describe the bug

If I try to embed a video like this "<iframe src="https://piped.video/embed/[id video]"></iframe>", the video is shown as intended but scrolling is enabled and is possible to view the footer of the page by scrolling down.
But this is not true if I view the video with the same link on a new page.

### To Reproduce

1. Open a simple html with the video as source of an iframe
2. Go on top of where the ifram is located
3. Use the scroll wheel to go down

### Expected behavior

Not being able to scroll the page or not being able to see the footer of the page

### Logs/Errors

Don't have any

Firefox 113.0.1 (64-bit) Arch Linux

### Additional context

_No response_

Bnyro:
Sometimes the footer and the navbar can be seen by scrolling for embedded videos, which should be fixed by that.

closes #2438


Bnyro:
I don't really like the approach to access the root `html` element from the `WatchVideo` component, but I couldn't find a better way to dynamically update the style properties of the `html` root element using Vue.

FireMasterK (approved):

FireMasterK:
> I don't really like the approach to access the root `html` element from the `WatchVideo` component, but I couldn't find a better way to dynamically update the style properties of the `html` root element using Vue.

It's not too bad, but I couldn't think of anything better :p

PR #2684, opened 2023-07-19
Change video bar colours to be defined as css vars.

amogusussy:
I like to customize websites I use a lot, and the only feature that I can't customize using CSS with piped is the progress bar, because it updates automatically, and it's defined as a linear gradient, so you can't redefine it or else the bar wont move.
This PR changes the colours to be defined as CSS vars, rather than just definite values. This makes styling it possible.

This is my first time doing anything with JavaScript (beside small user scripts), so please tell me if the way this is implemented isn't considered best practice.

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -479,12 +479,16 @@ export default {
                     overflowMenuButtons.push("open_new_tab");
                 }
 
+                document.documentElement.style.setProperty('--player-base', 'rgba(255, 255, 255, 0.3)');
+                document.documentElement.style.setProperty('--player-buffered', 'rgba(255, 255, 255, 0.54)');
+                document.documentElement.style.setProperty('--player-played', 'rgb(255, 0, 0)');
+
                 const config = {
                     overflowMenuButtons: overflowMenuButtons,
                     seekBarColors: {
-                        base: "rgba(255, 255, 255, 0.3)",
-                        buffered: "rgba(255, 255, 255, 0.54)",
-                        played: "rgb(255, 0, 0)",
+                        base: "var(--player-base)",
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer.vue
```diff
@@ -622,34 +639,17 @@ export default {
...
                     case "filler":
-                        color = "#7300FF";
-                        break;
                     case "music_offtopic":
-                        color = "#ff9900";
-                        break;
+                        color = "var(--spon-seg-${segment.category})"
                     default:
-                        color = "white";
+                        color = "var(--spon-seg-default)";
```
Bnyro, line 652: This should be sth like
```javascript
const color = ["sponsor", "selfpromo", ...].includes(segment.category) ? "var(--spon-seg-${segment.category})" : "var(--spon-seg-default)";
```

Bnyro:
And all the color variables should be defined via CSS, and not via JavaScript.

amogusussy:
> And all the color variables should be defined via CSS, and not via JavaScript.

How do I do that? There aren't any .css files in the project.

src/components/VideoPlayer.vue
```diff
@@ -479,12 +479,16 @@ export default {
                     overflowMenuButtons.push("open_new_tab");
                 }
 
+                document.documentElement.style.setProperty('--player-base', 'rgba(255, 255, 255, 0.3)');
+                document.documentElement.style.setProperty('--player-buffered', 'rgba(255, 255, 255, 0.54)');
+                document.documentElement.style.setProperty('--player-played', 'rgb(255, 0, 0)');
```
FireMasterK, line 484: Style shouldn't be set here, it should ideally be in the `<style>` tags.

amogusussy:
#2709

PR #2729, opened 2023-07-30
Use var() for video player's progress and audio bar.

amogusussy:
I like to customize websites I use a lot, and the only feature that I can't customize using CSS with piped is the progress bar, because it updates automatically, and it's defined as a linear gradient, so you can't redefine it or else the bar wont move.
This PR changes the colours to be defined as CSS vars, rather than just definite values. This makes styling it possible.

This is my first time doing anything with JavaScript (beside small user scripts), so please tell me if the way this is implemented isn't considered best practice.

#2684

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -482,9 +482,9 @@ export default {
                 const config = {
                     overflowMenuButtons: overflowMenuButtons,
                     seekBarColors: {
-                        base: "rgba(255, 255, 255, 0.3)",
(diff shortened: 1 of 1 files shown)


amogusussy:
I haven't done it for the volume bar yet, since I don't know where it's defined in the code.

Bnyro (approved):

Bnyro:
We currently don't override the color of the volume bar, you'll have to inspect it with the dev tools to figure out its class, so that you can change its color.

PR #2744, opened 2023-08-01
Fix: FreeTube Subscription Import

ChunkyProgrammer:
This PR fixes importing FreeTube subscriptions for people with more than one profile.

### DIFF
--- a/src/components/ImportPage.vue
+++ b/src/components/ImportPage.vue
@@ -106,13 +106,12 @@ export default {
                 // FreeTube DB
                 else if (text.indexOf("allChannels") != -1) {
                     const lines = text.split("\n");
-                    for (let i = 0; i < lines.length; i++) {
-                        if (lines[i] !== "") {
-                            const json = JSON.parse(lines[i]);
-                            json.subscriptions.forEach(item => {
-                                this.subscriptions.push(item.id);
-                            });
-                        }
+                    for (let line of lines) {
+                        if (line === "") continue;
+                        const json = JSON.parse(line);
+                        json.subscriptions.forEach(item => {
+                            this.subscriptions.push(item.id);
+                        });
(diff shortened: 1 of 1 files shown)


Bnyro:
This assumes there's one json "file" per line, right?
I think this could break the import feature for some users because not everyone's aware they need to keep the json per profile in one line.
At FreeTube, do you import the subscriptions line per line as well?

Bnyro:
I think that it'd make sense to do it like that if multiple different profiles are contained in the file,
however it'd probably also make sense to have some sort of fallback if there's only one profile, but at multiple lines.

ChunkyProgrammer:
For FreeTube we also import the subscriptions per line. 
Ref: https://github.com/FreeTubeApp/FreeTube/blob/2296e3a58d1895556bcec52b2162861140f295f8/src/renderer/components/data-settings/data-settings.js#L124

FireMasterK (approved):

src/components/ImportPage.vue
```diff
@@ -105,10 +105,15 @@ export default {
...
-                    });
+                    const lines = text.split("\n");
+                    for (let i = 0; i < lines.length; i++) {
+                        if (lines[i] !== "") {
+                            const json = JSON.parse(lines[i]);
+                            json.subscriptions.forEach(item => {
+                                this.subscriptions.push(item.id);
+                            });
+                        }
+                    }
```
Bnyro, line 116:
```suggestion
                    const lines = text.split("\n").filter(line => line !== "");
                    for (let line of lines) {
                        const json = JSON.parse(line);
                        json.subscriptions.forEach(item => {
                            this.subscriptions.push(item.id);
                        });
                    }
```

Bnyro (approved):

PR #2765, opened 2023-08-04
feat: qrcode share option

### LINKED ISSUE #2730: QR code "Share" option for switching to mobile.
### Describe the feature

In the "share" menu, add a QR code with the link to the video with the timestamp (if specified).

This would make it easier to switch to a mobile device when watching on the computer.

### Concept(s)

There'd be an auto-updating (according to settings) QR code in the "share" menu.

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #2730

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
         "hotkeys-js": "3.12.0",
         "javascript-time-ago": "2.5.9",
         "mux.js": "6.3.0",
+        "qrcode": "^1.5.3",
         "shaka-player": "4.3.8",
         "stream-browserify": "3.0.0",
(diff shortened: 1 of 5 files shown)


Bnyro:
Can't we only import the qr code dependency when needed?
I don't think a dedicated component for a simple canvas would be a good idea in general.

FireMasterK:
> Can't we only import the qr code dependency when needed?
> I don't think a dedicated component for a simple canvas would be a good idea in general.

I think the imports have to be a top level statement, which is why we need to do that.

PR #2806, opened 2023-08-14
EFY: Latest version Update

dragos-efy:
It's still the old version of Piped, but the new one will come up soon. Now you have:

- 1-12 main colors
- more efficiency and quicker translations
- notification control (position and type)
- support for transparent windows and video wallpapers (but those won't save yet on page reload)
- bigger ranges for visual filters
- bug fixes and more


Bnyro:
The linter complaints are just the ones as usual for efy and can be ignored.

dragos-efy:
well `extra.js` is not loaded so that should be fixed (I'll do it very soon), but otherwise yeah, the `extra_local.js` complaints don't make sense since they're server based, but the file is supposed to be used locally

PR #2857, opened 2023-08-24
Adding padding to center search icon in navbar

arisuv:
The search icon was not aligned in the center so I added even alignment on the x-axis. It will nicely align all inputs since the change on the input class.


FireMasterK (approved):

Bnyro:
Thanks!

PR #2879, opened 2023-08-30
Display error in player

### LINKED ISSUE #2836: Show a user friendly error message when a video can't be loaded
### Describe the feature

Currently, when a video can't be loaded (e.g. browser console or network requests show 403, which I assume means YouTube refuses to provide the video to the Piped instance or the Piped instance refuses to provide the video to my computer) nothing happens when clicking on the player, so a user has no idea what's going on. Pressing spacebar changes the play button into a pause button, but nothing else happens either. This is very frustrating, as we have no idea what's going on and if we can do anything to fix it. If an informative error message would be displayed in the player instead, people the world over would rejoice, for they'd have been imbued with knowledge instead of being left to stumble in the dark.

This would make it easier for the user to understand what's going on, and if the problem is on their end or not.

### Concept(s)

piped user friendly error message

### Additional context

_No response_

### Acknowledgements

snematoda:
closes #2836

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -25,6 +25,11 @@
             <span v-t="'actions.skip_segment'" />
             <i class="material-icons-round">skip_next</i>
         </button>
+        <span
+            v-if="error > 0"
+            v-t="{ path: 'player.failed', args: [error] }"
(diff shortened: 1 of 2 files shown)


FireMasterK (approved):
Thank you!

Issue #2904, opened 2023-09-04
Introduce a config file and/or env variable for default preferences of self-hosted instances


kovdan01:
### Describe the feature

When self-hosting an instance via docker (https://github.com/TeamPiped/Piped-Docker), a person gets an instance with default settings which can not be easily changed. It means that they are saved on the client side via cookies, which, for example, leads to using the undesired preferences in private browser windows and any new browser opened. While URL parameters exists and it might be possible to implicitly pass them (see a related question https://github.com/TeamPiped/Piped/issues/2903), it's definitely not a clean way for doing that. Having some kind of config file or environment variable for each of the settings on the preferences page would make changing default instance preferences way easier.

### Why would this be useful to add?

A person hosting the instance might have another views on which features are useful and which are not - for example, one might want to minimize description, comments and chapters by default and increase the buffering goal.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

PR #2916, opened 2023-09-07
Add config for privacy policy

wcw2x:
The option for adding a privacy policy link is missing, so I tried to implement it. **I don't have any experience with Vue, please take a double look at what I did.**

It's not possible to edit the default config (#2904), but it could be used in a future update.


FireMasterK (approved):
LGTM, thank you!

Btw, you can set the frontend property by using this in the backend configuration:
```
frontend.privacyPolicyUrl: https://your-url
```

wcw2x:
Good to know! Hope everyone will make a good use of it then :)

PR #2934, opened 2023-09-11
Big Piped + EFY Update

dragos-efy:
Fix lots of merge conflicts from master branch to efy, visual glitches and add improvements


Bnyro:
Looks like the CI can't build it because it doesn't clone the efy submodule? Didn't we change that once? :/

dragos-efy:
Technically yes lol. But I updated it so maybe that did something?

Issue #2669, opened 2023-07-15
URL-based preferences


syboxez:
### Describe the feature

Option to save preferences as a bookmark in addition to using cookies. The URL would save a cookie with the settings specified by the URL, that way not every URL would need to have all of the arguments.

### Why would this be useful to add?

For browsers that are always in private mode, it would be nice to be able to bookmark a URL with all of my preferences saved.

### Concept(s)

Example: `https://piped.video/?autoplay=false&proxy_lbry=true&enabled_codecs=av1,vp9,avc&default_quality=4320` would be set as a browser bookmark, and even if cookies are cleared on close (like in private browsing), the settings would still be saved, and there would be no need to restore preferences on every browser launch.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

kaoneko:
If your browser which is always in Private Mode (or has its site data cleared regularly) runs uBlock Origin, you can set default preferences by adding filters like these:

`piped.video,kavin.rocks,whatever-instances-you-u.se##+js(trusted-set-local-storage-item, quality, 1080)`

You can find the key-value pairs by inspecting your browsers local storage through its developer tools after configuring your preferred settings. The advantage of this approach is that you are not limited to options supported by URL parameters and you don't have to do any URL rewriting either.

kaoneko:
You have to use one line per setting. Also, I just discovered that per today's update of uBO you can't just use `trusted-set-local-storage-item` in your personal filters without changing an advanced setting first. Settings that use `true`, `false` or a number as the value can be set with `set-local-storage-item` though. My uBO filters for Piped settings look like this:
```
piped.video##+js(set-local-storage-item, dearrow, true)
piped.video##+js(trusted-set-local-storage-item, enabledCodecs, avc)
piped.video##+js(set-local-storage-item, minimizeComments, true)
piped.video##+js(set-local-storage-item, minimizeRecommendations, true)
piped.video##+js(set-local-storage-item, playerAutoPlay, false)
piped.video##+js(set-local-storage-item, quality, 1080)
piped.video##+js(set-local-storage-item, showWatchOnYouTube, true)
```
[...]

Issue #223, opened 2021-06-24
Allow preferences to be "set" from query parameters.


FireMasterK:
and take Invidious's naming scheme wherever possible.

This way users without cookies enabled have a workaround.

FireMasterK:
Partly done in 4a10d80804105d50cf0d944d537f4482051b0fe5.

FireMasterK:
Closing this with 7a77534098d2a19ad4c67e58ecb30dcea778776d.

Piped should without cookies theoretically now.

PR #2935, opened 2023-09-11
Allow set subtitle language from url parameter

aiosk:
**As** user,
**I want to** set subtitles language from url parameters, 
**So i** do not need to set subtitle language each time i switch instances

you can set `subtitles=xx` (example: `subtitles=en`) on url parameter to set video subtitle language. full url example: https://piped.video/watch?v=6stlCkUDG_s&subtitles=en

i choose query `subtitles` so it's compatible with [invidious url parameter](https://docs.invidious.io/url-parameters/)

this is especially useful when you have addons like redirector

improvement over https://github.com/TeamPiped/Piped/issues/2669 and https://github.com/TeamPiped/Piped/issues/223

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -594,6 +594,14 @@ export default {
 
                     const autoDisplayCaptions = this.getPreferenceBoolean("autoDisplayCaptions", false);
                     this.$player.setTextTrackVisibility(autoDisplayCaptions);
+
+                    const prefSubtitles = this.getPreferenceString("subtitles", null);
+                    const textTracks = this.$player.getTextTracks();
+                    const subtitleIdx = textTracks.findIndex(textTrack => textTrack.language == prefSubtitles);
+                    if (subtitleIdx >= 0) {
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer.vue
```diff
@@ -594,6 +594,18 @@ export default {
...
                     const autoDisplayCaptions = this.getPreferenceBoolean("autoDisplayCaptions", false);
                     this.$player.setTextTrackVisibility(autoDisplayCaptions);
+
+                    (() => {
+                        const prefSubtitles = this.getPreferenceString("subtitles", "");
+                        if (prefSubtitles === "") return;
+
+                        const textTracks = this.$player.getTextTracks();
+                        const subtitleIdx = textTracks.findIndex(textTrack => textTrack.language == prefSubtitles);
+                        if (subtitleIdx == -1) return;
```
FireMasterK, line 604: Instead of creating a function and returning, can we just use `if (subtitleIdx != -1) {}`?

Bnyro:
Thatnks!

Issue #1761, opened 2022-11-16
Don't show audio language option if only one language exists


AlyoshaVasilieva:
### Describe the feature

When there's only one language for a video, don't display the "Language" option in the meatball menu.

### Why would this be useful to add?

As it exists it's mostly pointless for single-language videos, since they seem to mostly list their language as "Undetermined", so it doesn't present any useful info.

For embedded videos, the Language menu item means that the "Open in new tab" menu item cannot be seen until you scroll; this item is important for embeds since they're typically quite small. (note: only for some embeds, since not all embeds have e.g. captions)

![embedded video with menu open](https://user-images.githubusercontent.com/1284317/202295733-988b0c5d-c66e-4059-ab06-9c474ac2c932.png)

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

PR #2937, opened 2023-09-12
Don't show audio language option if language is undetermined

aiosk:
as a viewer
i want simple option
so i
  - don't need to see unnecessary option
  - have more simple option 

by not showing audio language option if language is undetermined

fix https://github.com/TeamPiped/Piped/issues/1761

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -537,10 +537,8 @@ export default {
                         player.selectAudioLanguage(lang);
                     }
 
-                    (() => {
-                        const audioLanguages = player.getAudioLanguages();
-                        if (audioLanguages.length == 1 && audioLanguages[0] == "und") return;
-
+                    const audioLanguages = player.getAudioLanguages();
+                    if (audioLanguages.length > 1) {
                         const overflowMenuButtons = this.$ui.getConfiguration().overflowMenuButtons;
                         // append language menu on index 1
                         const newOverflowMenuButtons = [
@@ -549,7 +547,7 @@ export default {
                             ...overflowMenuButtons.slice(1),
                         ];
                         this.$ui.configure("overflowMenuButtons", newOverflowMenuButtons);
-                    })();
+                    }
 
                     if (qualityConds) {
                         var leastDiff = Number.MAX_VALUE;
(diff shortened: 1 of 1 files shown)


src/components/VideoPlayer.vue
```diff
@@ -544,6 +537,20 @@ export default {
                         player.selectAudioLanguage(lang);
                     }
 
+                    (() => {
```
Bnyro, line 540: Why are we creating a function here and call it immediately after?

aiosk: it's called [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).
i used it to [Avoid polluting the global namespace](https://developer.mozilla.org/en-US/docs/Glossary/IIFE#avoid_polluting_the_global_namespace) and [Reduce Nesting, Reduce Complexity](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html).
in this case, to [reduce nested if](https://www.jetbrains.com/help/resharper/InvertIf.html)

without IIFE, the line `if (audioLanguages.length == 1 && audioLanguages[0] == "und") return;`  will exit code from function `setPlayerAttrs()` and will not execute the next code

without IIFE and attempt to reduce nesting, the code would look like this
```js
const audioLanguages = player.getAudioLanguages();
if (audioLanguages.length > 1) {
  const overflowMenuButtons = this.$ui.getConfiguration().overflowMenuButtons;
  // append language menu on index 1
  const newOverflowMenuButtons = [
      ...overflowMenuButtons.slice(0, 1),
      "language",
      ...overflowMenuButtons.slice(1),
  ];
  this.$ui.configure("overflowMenuButtons", newOverflowMenuButtons);
}
```

it's your call. i will make the change
i just want this feature..

Bnyro: Thanks for the explanation!

In my opinion using an if statement looks simpler and more self-explaining here. "IIFE" is certainly a pretty cool thing, but only few people are familiar with and our goal is to make everything as easy to maintain and understand as possible.

Hence it would be great if you could replace the "IIFE" with an if loop (also in your other PR), then I'd be very glad to merge both of them (or Kavin) :)

aiosk: solved by https://github.com/TeamPiped/Piped/pull/2937/commits/9b740676acca541fc4432e932d02e047382b604b

Bnyro (approved):
Thanks!

PR #2950, opened 2023-09-14
feat: add to channel group modal on channel page

### LINKED ISSUE #2922: Add 'Add to Group' button
### Describe the feature

With the frequent addition of many new channels to subscriptions, going to the group settings each time is cumbersome.
This could be solved by adding an `Add to Group` button next to the existing one.

It would be convenient to be able to add a channel to a group directly after hitting `Subscribe`.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #2922

### DIFF
--- a/src/components/ChannelPage.vue
+++ b/src/components/ChannelPage.vue
@@ -27,6 +27,13 @@
                     @click="subscribeHandler"
                 ></button>
 
+                <button
+                    v-if="subscribed"
+                    v-t="'actions.add_to_group'"
+                    class="btn"
+                    @click="showGroupModal = true"
+                ></button>
+
                 <!-- RSS Feed button -->
                 <a
                     v-if="channel.id"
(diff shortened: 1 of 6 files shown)


src/components/ChannelPage.vue
```diff
@@ -27,6 +27,13 @@
                     @click="subscribeHandler"
                 ></button>
 
+                <button
+                    v-if="subscribed"
+                    v-t="'actions.add_to_group'"
```
FireMasterK, line 32: Should this be manage group instead, considering you could technically use this to remove it from a channel group too?

Bnyro: Probably something like "Change group membership", but I think  "Add to group" is just more self-explaining (and shorter) in that case.

"Manage group" sounds wrong because you don't manage a group, you change whether the channel is part of the groups.

PR #2951, opened 2023-09-14
feat: show explaination on register and login page

### LINKED ISSUE #1288: More information on login and register page
### Describe the feature

On the register page it is not clear what the registration means and why you should register.

Why should someone register an account? Does it allow you to like/dislike video, comment on videos, add videos to a playlist, star or bookmark videos? What are the benefits?

Does it create a YouTube account, or is it a local Piped account, or a global account that can be used on any Piped instance?

When you login can you use your YouTube account to login? Can you use an account registered on another Piped instance?
What if I forgot my password?

To make it easier to understand what the login and registration means, how it works, what benefits it provide.

### Concept(s)

_No response_

### Additional context

_No response_

### Acknowledgements

Bnyro:
closes #1288


FireMasterK:
Could we maybe have an icon like https://fontawesome.com/v4/icon/info-circle and show this information on hover or tap instead?

Bnyro:
Good idea :+1:

vanillajonathan:
The comment about choosing a strong password is rather superfluous.

PR #3010, opened 2023-10-07
Use instance name when downloading playlist as .txt

amogusussy:
Right now, when you download a playlist .txt file, it uses "piped.video". This PR makes it use the current instance name.


Bnyro:
This behavior has been intended, yt-dlp does not support all Piped instance domains, so it makes sense to only use "piped.video" to make sure yt-dlp can download the playlist file.

PR #3014, opened 2023-10-11
Set fixed uploader avatar size on Watch Video page

GlenLowland:
In new backend version the avatar sizes became too big. Fix it by setting the fixed image size.


FireMasterK (approved):
Thank you!

Issue #3070, opened 2023-10-26
How to simply play a playlist?


ribtoks:
### Describe your question

When I select any video from the playlist, it does not play any other videos, but just the one I clicked, like it is completely separate. Is there a possibility to play all videos from the playlist?

The link, that is opened, does contain `list=` argument, so it looks like the one on YouTube.

FireMasterK:
Yes, enable the "Auto Play next Video" option on the same page.

PR #3088, opened 2023-10-30
Minimize the video description by default

robertgro:
please for the love of the user experience https://github.com/TeamPiped/Piped/issues/3070

### DIFF
--- a/src/components/PreferencesPage.vue
+++ b/src/components/PreferencesPage.vue
@@ -417,7 +417,7 @@ export default {
             countrySelected: "US",
             defaultHomepage: "trending",
             minimizeComments: false,
-            minimizeDescription: false,
+            minimizeDescription: true,
             minimizeRecommendations: false,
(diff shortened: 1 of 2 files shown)


Bnyro:
In my opinion we should minimize the video description by default instead (even though there's already a preference for it). Moving the buttons up like that looks kinda out of place.

robertgro:
Maybe this is what is what you looking for @Bnyro.

Bnyro:
It would need to be minimzed at `src/components/PreferencesPage.vue` too by default.

Opinion @FireMasterK?

robertgro:
> It would need to be minimzed at `src/components/PreferencesPage.vue` too by default.

This should do the trick.

Bnyro (approved):

Bnyro (approved):
Thanks!

robertgro:
You are welcome.

PR #3093, opened 2023-10-30
Update package.json

maathimself:
Downgrade shaka-player to 4.3.9 to fix live stream issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
         "linkifyjs": "4.1.1",
         "mux.js": "6.3.0",
         "qrcode": "^1.5.3",
-        "shaka-player": "4.5.0",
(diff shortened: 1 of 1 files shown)


maathimself:
If someone could help me get this to work that would be great. I don't know what I'm doing so it can also just be removed.

Bnyro:
`pnpm install shaka-player@4.3.9` should work I think.

Bnyro:
We can also just revert the dependency bump via git. I'm just not sure if that's really the cause of the issue. Any upstream reference issue at Shaka?

np22-jpg:
> Downgrade shaka-player to 4.3.9 to fix live stream issue

Is there a reference issue documenting the live streaming problems?

PR #3127, opened 2023-11-09
add vidyodl to list of Made with Piped in README

MrKovar:
I recently finished up a project intended for use on my home server that I figured this group of devs may find either useful or want to use themselves. It uses the list of proxies from piped-proxy repo to find whichever proxy is up and then determines the fastest one for a quick video downloading process using ffmpeg.


FireMasterK (approved):
Interesting project, thank you!

PR #3166, opened 2023-11-21
Dockerfile: rm EXPOSE

kskarthik:
The EXPOSE is inferred from the base image which in this case is nginx

So, we need not expicitly define it. One less layer in final image :)


FireMasterK (approved):
Thanks!

PR #3174, opened 2023-11-24
chore: re-structure projects using Piped in README.md

maboroshin:
The apps list is cluttered.
* Separate apps for OS and online services
* Grouping for android, video, music, etc. (Simple continuity. Strict rules are difficult.)
* Move up multi-language apps
* Move up starred apps

----

PlasmaTube for Linux seems to be in the process of supporting Piped. This could be the development version. Add to the list now? Or wait until the official version is released?
https://invent.kde.org/multimedia/plasmatube#supported-sources

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,19 +142,22 @@ Contributions in any other form are also welcomed.
 
 # Made with Piped
 
--   [Yattee](https://github.com/yattee/yattee) - An alternative frontend for YouTube, for IOS.
--   [LibreTube](https://github.com/Libre-tube/LibreTube) - An alternative frontend for YouTube, for Android.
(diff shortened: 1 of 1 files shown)


Bnyro (changes requested):
See above

maboroshin:
I changed to those categories.

PlasmaTube seems to support Piped in version 23.08.3 (2023-11-09). https://apps.kde.org/plasmatube/ And when I try to get it from FlatHub, it is still the previous version 23.08.2 (2023-10-12). So I have not tested 23.08.3. I have not tried it because I using Live OS by DVD. But It will be easily available from FlatHub in the near future.

Bnyro (approved):
Lgtm, Thanks!

PR #3181, opened 2023-11-26
deps: remove mux.js

snematoda:
mux.js seems to be depreciated since shaka 4.5

https://github.com/shaka-project/shaka-player/blob/2a524bf51fc613b8ebddb5794524cbb3f6366d4b/lib/dependencies/all.js#L28


FireMasterK (approved):
Nice spot, thank you!

PR #3264, opened 2023-12-18
Fix image lazy loading for Firefox

anewuser:
Hello. Firefox is requesting images that should be lazy-loaded. To reproduce:

1. Set Piped to minimize recommendations by default
2. Open https://piped.video/watch?v=jNQXAC9IVRw and monitor your network requests

Due to [Bug 1647077](https://bugzilla.mozilla.org/show_bug.cgi?id=1647077), we need to place `loading="lazy"` before the `src` attribute in `` tags for it to work properly.


Bnyro (approved):
Seems reasonable, thank you!

PR #3284, opened 2023-12-26
Allow setting API, proxy, and frontend URLs at build time + support subpaths

fkrull:
I guess this is sorta two features in one:
* I replaced most occurrences of hardcoded URLs in the code with references to environment variables that can be set at build time.
  - I left out opensearch.xml because I think expanding variables in that file would've required an additional vite plugin and I didn't want to mess with the build to that degree.
* Make it so vite's base setting can be used to run the frontend in a subpath rather than on the toplevel of a domain. This one is a bit of a best effort: I tested it and everything I tried worked, but I'm not sure I fixed every link that might ignore the subpath. But the changes shouldn't (fingers crossed) make any difference for the default "/" base URL.

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,3 @@
+VITE_PIPED_API=https://pipedapi.kavin.rocks
+VITE_PIPED_PROXY=https://pipedproxy.kavin.rocks
+VITE_PIPED_INSTANCES=https://piped-instances.kavin.rocks/
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
 VITE_PIPED_API=https://pipedapi.kavin.rocks
 VITE_PIPED_PROXY=https://pipedproxy.kavin.rocks
 VITE_PIPED_INSTANCES=https://piped-instances.kavin.rocks/
+VITE_PIPED_FRONTEND=https://piped.video
 VITE_BASE_URL=/
--- a/.env
+++ b/.env
@@ -1,4 +1,3 @@
 VITE_PIPED_API=https://pipedapi.kavin.rocks
 VITE_PIPED_PROXY=https://pipedproxy.kavin.rocks
 VITE_PIPED_INSTANCES=https://piped-instances.kavin.rocks/
-VITE_BASE_URL=/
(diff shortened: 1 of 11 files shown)


src/components/PlaylistPage.vue
```diff
@@ -158,7 +158,7 @@ export default {
         downloadPlaylistAsTxt() {
             var data = "";
             this.playlist.relatedStreams.forEach(element => {
-                data += "https://piped.video" + element.url + "\n";
+                data += import.meta.env.VITE_PIPED_FRONTEND + element.url + "\n";
```
FireMasterK, line 161: I'm not so sure that this should be changed, this could break link discovery for things like `yt-dlp`, etc.

FireMasterK (approved):

src/components/PlaylistPage.vue
fkrull: Oh, as in tools like that won't discover URLs because they won't be piped.video any more for deployments on other domains? I suppose it's possible, but in turn it's not great when playlists send you to an entirely different domain either.

I'm not super married to this particular change, so I'll revert it if you prefer.

.env
```diff
@@ -0,0 +1,4 @@
+VITE_PIPED_API=https://pipedapi.kavin.rocks
+VITE_PIPED_PROXY=https://pipedproxy.kavin.rocks
+VITE_PIPED_INSTANCES=https://piped-instances.kavin.rocks/
+VITE_BASE_URL=/
```
Bnyro, line 4: I think `VITE_BASE_PATH` is more appropriate for this env variable.

fkrull: So I realised that setting the vite base path from this env file doesn't actually work. Apparently variables loaded from env files aren't available yet in the config file. However, I found out you can do `vite build --base=/whatever` to override the base setting which is even better than an environment variable.

The uses in the code itself have to stay `BASE_URL` because that's the variable vite exports to access its base path: https://vitejs.dev/guide/build#public-base-path

Bnyro (approved):
Thanks :+1:

PR #3325, opened 2024-01-11
fix: subscription groups use invalid channel id length

Bnyro:
**WHOOPS**

Merging this would break compatibility with current channel groups however, not sure what the best way to work around this.

(Noticed when working on import/export compatibility with LibreTube.)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -36,7 +36,7 @@ export default {
         });
 
         if ("indexedDB" in window) {
-            const request = indexedDB.open("piped-db", 5);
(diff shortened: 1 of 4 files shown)


FireMasterK:
I think we can add an IndexedDB migration to fetch the user's subscriptions, and then find the appropriate 24 character channel id from it.

Bnyro:
Right, good idea :+1:
I'll look into this tomorrow.

Bnyro:
I've added some migrations now, and tested with a few channels and groups and seems to be working fine, the database is being updated properly.

PR #3346, opened 2024-01-19
feat: support for subscribing to channels in search results

### LINKED ISSUE #509: Quick Subscribe: Showing 'Subscribe' button on search results page
Hello,

It'd be great if I could see the Subscribe button displayed on the search results page in the same or similar way Youtube does.

image

Thank you!

Bnyro:
closes #509

### DIFF
--- a/src/components/ChannelItem.vue
+++ b/src/components/ChannelItem.vue
@@ -1,37 +1,72 @@
 <template>
     <div class="flex flex-col flex-justify-between">
-        <router-link :to="props.item.url">
+        <router-link :to="item.url">
             <div class="my-4 flex justify-center">
-                <img loading="lazy" class="aspect-square w-[50%] rounded-full" :src="props.item.thumbnail" />
+                <img loading="lazy" class="aspect-square w-[50%] rounded-full" :src="item.thumbnail" />
             </div>
             <p>
-                <span v-text="props.item.name" />
-                <font-awesome-icon v-if="props.item.verified" class="ml-1.5" icon="check" />
+                <span v-text="item.name" />
+                <font-awesome-icon v-if="item.verified" class="ml-1.5" icon="check" />
             </p>
         </router-link>
-        <p v-if="props.item.description" v-text="props.item.description" />
-        <router-link v-if="props.item.uploaderUrl" class="link" :to="props.item.uploaderUrl">
+        <p v-if="item.description" v-text="item.description" />
+        <router-link v-if="item.uploaderUrl" class="link" :to="item.uploaderUrl">
             <p>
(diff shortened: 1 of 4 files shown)


src/components/ChannelItem.vue
```diff
@@ -1,37 +1,72 @@
...
+            this.subscribed = await this.fetchSubscriptionStatus(this.channelId);
+            console.log(this.subscribed);
+        },
+        subscribeHandler() {
+            this.toggleSubscriptionState(this.channelId, this.subscribed).then(success => {
+                if (success) this.subscribed = !this.subscribed;
+            });
+        },
     },
-});
```
FireMasterK, line 36: We should not change this to the legacy Options API, we should be using the [Composition API](https://vuejs.org/guide/extras/composition-api-faq.html) like before ideally.

Ideally, we should migrate our existing code to the composition api.

FireMasterK (approved):

src/components/ChannelItem.vue
Bnyro: Looks awesome, quite similar to Svelte and a lot less boilerplate code :)

Bnyro: How should we be using `this` now? The Vue docs say that we should migrate to replacing mixins with [Composable functions](https://vuejs.org/guide/reusability/composables), but that's certainly out of scope for this PR.

src/components/ChannelItem.vue
```diff
@@ -23,15 +23,42 @@
...
+const channelId = props.item.url.substr(-24);
+
+onMounted(() => {
+    updateSubscribedStatus();
+});
+async function updateSubscribedStatus() {
+    subscribed.value = await this.fetchSubscriptionStatus(channelId);
+}
+function subscribeHandler() {
+    this.toggleSubscriptionState(channelId, subscribed.value).then(success => {
```
Bnyro, line 60: `this` not found

src/components/ChannelItem.vue
```diff
@@ -23,15 +23,42 @@
...
     },
 });
+const subscribed = ref(null);
+const channelId = props.item.url.substr(-24);
+
+onMounted(() => {
+    updateSubscribedStatus();
+});
+async function updateSubscribedStatus() {
+    subscribed.value = await this.fetchSubscriptionStatus(channelId);
```
Bnyro, line 57: `this` not found

src/components/ChannelItem.vue
Bnyro: Continuing the use of the composition API would require major refactors, which I don't intend to do with this PR.

PR #3357, opened 2024-01-23
fix: local playlists can't be auto-played

dadus33:
I believe this should fix my previously reported bug #3356 but even though it's a tiny change, I have zero knowledge of frontends in general, even more so Vue, so please review carefully for any potential side effects I may have missed.

closes #3356

PR #3361, opened 2024-01-25
fix: API retrieved playlists not shown in watch video page

dadus33:
My previous PR (#3357), although it fixed local playlists being shown in the watch page, it apparently broke API retrieved playlists (both YouTube playlists as well as those saved on instances when authenticated), hence this PR that aims to fix both - hopefully without introducing other bugs along the way, like my last PR 😅


FireMasterK (approved):
Thank you!

PR #3372, opened 2024-01-30
fix the alignment of tooltip in login & register pages

kskarthik:
image image

### DIFF
--- a/src/components/LoginPage.vue
+++ b/src/components/LoginPage.vue
@@ -1,5 +1,8 @@
 <template>
-    <h1 v-t="'titles.login'" class="my-4 text-center font-bold" />
+    <div class="flex justify-center">
+        <h1 v-t="'titles.login'" class="my-4 text-center font-bold" />
+        <TooltipIcon class="mb-6" icon="i-fa6-solid:circle-info" :tooltip="$t('info.login_note')" />
+    </div>
     <hr />
     <div class="w-full flex items-center justify-center text-center">
         <form class="w-min children:pb-3">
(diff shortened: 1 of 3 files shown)


Bnyro:
To be honest, I intentionally placed the tooltips where they are right now.

I'd prefer to keep them as they are, but it's a matter of personal taste.

FireMasterK:
<details>
<summary>
Current design:
</summary>

image
image
</details>

I personally think the information icon could utilize a better position, since currently it looks like it's in a random position. I do like the idea of how it's next to the button, like how this PR showcases. I'm open to other ideas too.

kskarthik:
> Maybe it makes the most sense to add it to the "Login" / "Register" title at the top of the page?
> 
> My idea would be to place it at the top right next to the title element.

I did also try that first. I felt apt beside the buttons. Let me revisit that idea again

Bnyro:
I'd place it more up so that it aligns with the top of the "Login" title, that looks a bit better in my opinion.

Bnyro:
Apart from that I prefer this one to the initial suggestion of this PR :+1:

kskarthik:
@Bnyro you meant this position?

image

Bnyro:
Yes, exactly :+1:

Bnyro (approved):

Bnyro:
(Approved once the changes are pushed)

kskarthik:
Done!

image 
image

Bnyro (approved):
Neat, thank you!

PR #3420, opened 2024-02-23
fix: Infinite loading when search query is empty

grittypuffy:
## Bug

When search query string is empty, Piped loads infinitely. This is replicable on hosted instances such as do.piped.video and in development environment as well.

Here's a screenshot for it (note the URL bar with null search_query):

URL bar showing empty search_query and the results are loaded infinitely

This is tested on

1. Chromium 121
2. Firefox 122

as well.

This PR aims to redirect users to landing page when the search query is empty (I am entirely not sure if a toast need to be showed for this to the user), and display results only when search query is non-null.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ pnpm install
 ### Compiles and hot-reloads for development
 
 ```
-pnpm serve
+pnpm dev
 ```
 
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Thank you!

PR #3430, opened 2024-02-27
feat: respect listen mode in recommended videos

Max-F-Helm:
When a video is watched as audio-only, this PR will affect the video-links (recommended videos and playlist videos) to open as audio-only per default.


src/components/VideoItem.vue
```diff
@@ -99,13 +100,13 @@
                             v: item.url.substr(-11),
                             ...(playlistId && { list: playlistId }),
                             ...(index >= 0 && { index: index + 1 }),
-                            listen: '1',
+                            ...(!preferListen && { listen: 1 }),
```
Bnyro, line 103: The negation here is a typo, or why would we use `!preferListen` here?

Bnyro (approved):
Looks good to me, thank you!

PR #3444, opened 2024-03-06
Replace font-awesome icons with unocss

### LINKED ISSUE #2718: Remove fortawesome dependency for unocss
I want to remove the fortawesome dependency entirely, as we have unocss which supports icons. We can then reduce our bundle size and dom element tree size (since the SVG isn't in the dom tree).

image

0x24D:
Closes #2718

### DIFF
--- a/src/components/WatchOnButton.vue
+++ b/src/components/WatchOnButton.vue
@@ -19,11 +19,13 @@ export default {
         <!-- For large screens -->
         <a :href="link" class="btn flex items-center lt-lg:hidden">
             <i18n-t keypath="player.watch_on" tag="strong">{{ platform }}</i18n-t>
-            <font-awesome-icon class="mx-1.5" :icon="['fab', platform.toLowerCase()]" />
+            <i v-if="platform == 'YouTube'" class="i-fa6-brands:youtube mx-1.5" />
+            <i v-else-if="platform == 'Odysee'" class="i-fa6-brands:odysee mx-1.5" />
         </a>
         <!-- For small screens -->
         <a :href="link" class="btn flex items-center lg:hidden">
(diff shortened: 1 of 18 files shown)


src/components/WatchOnButton.vue
```diff
@@ -19,11 +19,11 @@ export default {
         <!-- For large screens -->
         <a :href="link" class="btn flex items-center lt-lg:hidden">
             <i18n-t keypath="player.watch_on" tag="strong">{{ platform }}</i18n-t>
-            <font-awesome-icon class="mx-1.5" :icon="['fab', platform.toLowerCase()]" />
+            <i :class="'i-fa6-brands:' + platform.toLowerCase()" class="mx-1.5" />
```
0x24D, line 22: This is the only replacement that doesn't work and I'm unsure why.

1. i-fa6-brands:youtube is shown correctly in the DOM
2. Hardcoding this line to ```i-fa6-brands:youtube``` works correctly
3. Updating ```platform``` to another icon, i.e. Bitcoin, shows that icon correctly

0x24D: Actually, updating ```platform``` to Apple, or a brand that isn't already used, doesn't work either. I assume Bitcoin works because it is already cached from being used in the footer(?)

Bnyro: Yeah, the preprocessor probably only "compiles" the icons that it can find in the plain source code, and since we're using dynamic icon names here it can't recognize it. We should likely just do a `v-if / v-else` block here for each possible platform out of YouTube and Odysee.

Bnyro (approved):
Thank you!

PR #3475, opened 2024-03-16
feat: make `CreatePlaylistmodal` handle `Enter` key and auto focus input, `PlaylistAddModal` switches to created playlist

AndyRusso:
Hello.
This pull request has 4 changes:
- `CreatePlaylistModal` now handles `Enter` properly (creates the playlist and emits `"close"`)
- `PlaylistAddModal` used to add the video to playlist and close itself when pressing `Enter` while creating a playlist, this is fixed now
- `PlaylistAddModal` now switches to the created playlist
- while I was at it, I made `CreatePlaylistModal`'s input focus when mounted

It doesn't look like there is auto focus for inputs anywhere on the website, is this by design? aka should I revert the last commit?

The changes are broken into their own commits in the same order.

### DIFF
--- a/src/components/PlaylistAddModal.vue
+++ b/src/components/PlaylistAddModal.vue
@@ -65,7 +65,7 @@ export default {
     },
     methods: {
         handleKeyDown(event) {
-            if (event.code === "Enter") {
+            if (event.code === "Enter" && !this.showCreatePlaylistModal) {
                 this.handleClick(this.selectedPlaylist);
                 event.preventDefault();
             }
--- a/src/components/PlaylistAddModal.vue
+++ b/src/components/PlaylistAddModal.vue
@@ -22,7 +22,7 @@
     <CreatePlaylistModal
         v-if="showCreatePlaylistModal"
         @close="showCreatePlaylistModal = false"
-        @created="fetchPlaylists"
+        @created="addCreatedPlaylist"
     />
 </template>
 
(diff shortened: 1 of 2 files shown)


src/components/PlaylistAddModal.vue
```diff
@@ -87,7 +87,11 @@ export default {
                 if (json.error) alert(json.error);
             });
         },
-        async fetchPlaylists() {
+        async fetchPlaylists(createdPlaylistId) {
```
Bnyro, line 90: We should probably move the logic when a new playlist got created to a new function that
1. takes two arguments: `playlistId` and `playlistName`
2. appends the created playlist to the `playlists` array
3. sets the `selectedPlaylistId` to the `playlistId` of the new playlist

Bnyro: We don't actually need to do this API call here again when the user created a new playlist, since we have all info we need without any additional API request.

AndyRusso: Updated that locally for now, should I inline `fetchPlaylists` now that it's only used in `mounted`?

Bnyro: Yes, sounds reasonable :+1:

AndyRusso: Done!

Bnyro (approved):
Thank you! (just squashed some commits together to cleanup the history a bit)

AndyRusso:
Makes sense. Thanks!

PR #3482, opened 2024-03-18
fix: update text in search field when clicking on suggestion

guemax:
Currently, the text in the search field only updates itself when you select a suggestion via the up/down arrow keys, but _not_ when clicking on a suggestion with the left mouse button. However, the `h1` element below the input field changes in both cases. I personally find this quite irritating.

It seems like this is unintended behavior: There is a function named `onClick(i)` which does exactly the thing I would like to have, but apparently it is never called anywhere.

Thanks in advance for your answer!

### DIFF
--- a/src/components/SearchSuggestions.vue
+++ b/src/components/SearchSuggestions.vue
@@ -73,9 +73,6 @@ export default {
                 this.selected = i;
             }
         },
-        onClick(i) {
-            this.setSelected(i);
-        },
         setSelected(val) {
             this.selected = val;
             this.$emit("searchchange", this.searchSuggestions[this.selected]);
--- a/src/components/SearchSuggestions.vue
+++ b/src/components/SearchSuggestions.vue
@@ -1,7 +1,12 @@
 <template>
     <div class="suggestions-container absolute">
         <ul>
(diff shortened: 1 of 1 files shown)


src/components/SearchSuggestions.vue
```diff
@@ -1,7 +1,12 @@
 <template>
     <div class="suggestions-container absolute">
         <ul>
-            <li v-for="(suggestion, i) in searchSuggestions" :key="i" @mouseover="onMouseOver(i)">
+            <li
+                v-for="(suggestion, i) in searchSuggestions"
+                :key="i"
+                @mouseover="onMouseOver(i)"
+                @mousedown="onClick(i)"
```
Bnyro, line 8:
```suggestion
                @click="onClick(i)"
```
That's probably the more common way to do this?

Bnyro:
Thanks for your PR, what you said above absolutely makes sense :+1:

I think that the `onClick` method is not needed at all because we could just call the `setSelected` method directly instead, so I'd prefer just doing so as proposed above.

src/components/SearchSuggestions.vue
guemax: I've changed the code according to your suggestions. (Note: A quick `grep` showed that Piped never uses `@mousedown`, only `@click`. :wink: )

Bnyro (approved):

PR #3491, opened 2024-03-20
feat: eslint rule vue/no-undef-components makes build fail

Max-F-Helm:
To prevent mistakes like in #3490, I made the checking of eslint more strict.

Feel free to suggest more I should try to apply.


Bnyro (approved):
Thanks :+1:

PR #3509, opened 2024-03-25
chore(readme): add PlasmaTube for Linux

maboroshin:
PlasmaTube 24.02.1 for Linux is now available from flathub. https://flathub.org/apps/org.kde.plasmatube (At the time previously mentioned there was no Piped feature: https://github.com/TeamPiped/Piped/pull/3174#issuecomment-1836225682)

It's still experimental. Search in Piped is not working. To watch the video, you must press the play button. There is no login feature to Piped. But it may be implemented. because Invidious can log in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,10 @@ Contributions in any other form are also welcomed.
 # Made with Piped
 
 **Mobile/desktop apps**
--   [LibreTube](https://github.com/Libre-tube/LibreTube) - Alternative frontend for YouTube, for Android.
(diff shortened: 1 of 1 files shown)


Bnyro:
I think we should wait with this until search is possible via Piped.

Bnyro:
Looking at their source code, it seems that it does now support search via Piped (at least there's code for Piped's search API).

Bnyro (approved):
Thanks!

Issue #3579, opened 2024-04-29
Request confirmation before deleting subscription groups


abacabadabacaba:
### Describe the feature

Currently, subscription groups are deleted without confirmation. I propose to add a confirmation dialog.

### Why would this be useful to add?

This would prevent deleting subscription groups accidentally.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

khusroohayat:
[https://github.com/khusroohayat/Piped/pull/1](url)

Bnyro:
@khusroohayat You can just open a pull request at this repository here, you probably accidentally opened it at your own fork.

khusroohayat:
[https://github.com/TeamPiped/Piped/pull/3596](url) @Bnyro

PR #3596, opened 2024-05-07
feat: show confirmation modal before deleting subscription group

khusroohayat:
closes https://github.com/TeamPiped/Piped/issues/3579

### DIFF
--- a/src/components/SubscriptionsPage.vue
+++ b/src/components/SubscriptionsPage.vue
@@ -55,14 +55,13 @@
                 <i class="i-fa6-solid:pen mx-2" @click="showEditGroupModal = true" />
                 <i class="i-fa6-solid:circle-minus mx-2" @click="groupToDelete = group.groupName" />
             </div>
-            <ConfirmModal
-                v-if="groupToDelete == group.groupName"
-                :message="$t('actions.delete_group_confirm')"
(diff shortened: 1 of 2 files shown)


src/components/SubscriptionsPage.vue
```diff
@@ -53,9 +53,16 @@
             <span v-text="group.groupName !== '' ? group.groupName : $t('video.all')" />
             <div v-if="group.groupName != '' && selectedGroup == group">
                 <i class="i-fa6-solid:pen mx-2" @click="showEditGroupModal = true" />
-                <i class="i-fa6-solid:circle-minus mx-2" @click="deleteGroup(group)" />
+                <i class="i-fa6-solid:circle-minus mx-2" @click="groupToDelete = group.groupName" />
             </div>
+            <ConfirmModal
+                v-if="groupToDelete == group.groupName"
```
Bnyro, line 59: The modal could be placed outside the for loop with the condition `groupToDelete != null`, that way it's not unnecessarily doing that comparison for each group each time `groupToDelete` changes.

khusroohayat: @Bnyro done

Bnyro (approved):
Thank you!

PR #3621, opened 2024-05-16
fix: links in comments are not clickable

### LINKED ISSUE #1244: Make links in Comments sections clickable like a hyperlink.

Bnyro:
closes #1244


Bnyro:
> TFW a bug is fixed by deleting a bunch of lines of code...
>
Yeah, we had refactored the code to work better and fixed some issues, but apparently someone added some broken extra code only for comments in addition to the working one, so we parsed the comment text twice - first with the broken method - then with the working method (which doesn't work because everything got already broken by the first one, lol)

PR #3663, opened 2024-06-09
make the time in the video player preview white in light mode

ac615223s5:
image


Bnyro:
We need to use `.dark` instead of `dark` to also support when the user manually set the theme in the settings.

PR #3667, opened 2024-06-10
add a default name for new playlists

ac615223s5:
for creating playlists to use as a temporary \`queue`


Bnyro:
In my opinion there's no need to have a default playlist name - but if there's one, it should be unique, so that you can create multiple playlists quickly that all have different names in order to be distinguishable.

Bnyro:
Sorry, I don't really think that's the right way to solve that.

PR #3695, opened 2024-07-11
feat: add option to favor hls over dash

snematoda:
added an option to player preferences to out-out of dash and favor hls instead...


FireMasterK (approved):
Thanks!

PR #3701, opened 2024-07-12
Update search text and title when route changes

guemax:
Previously, only the `h1` below the search field updated when going back and forth in the browsing history. With this change, the title displayed in the tab and the text in the search field update as well, see the attached screen cast.

https://github.com/user-attachments/assets/d51bded1-447a-4688-9565-d9b662781a34


Bnyro (approved):
lgtm, thanks!

PR #3722, opened 2024-07-17
fix: duplicated time indicator when scrubbing seek bar

Bnyro:


Bnyro:
Caused by https://github.com/TeamPiped/Piped/pull/3601

PR #3806, opened 2024-10-02
feat: Run piped frontend with unprivileged user

bjw-s:
This PR allows running the Piped frontend container as an unprivileged user instead of running it with `root` permissions.

It also introduces two additional environment variables that allow some control over the number of nginx worker processes and the port nginx is exposed on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,11 +15,11 @@ RUN --mount=type=cache,target=/root/.local/share/pnpm \
     pnpm install --prefer-offline && \
     pnpm build && ./localizefonts.sh
 
-FROM nginx:alpine
+FROM nginxinc/nginx-unprivileged:alpine
 
-COPY --from=build /app/dist/ /usr/share/nginx/html/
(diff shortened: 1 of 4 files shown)


bjw-s:
@Bnyro / @FireMasterK as you appear to be the last non-bots to have merged any PR's, is there anything I can / need to do to progress this PR?

Bnyro (changes requested):
Please split the commit into two commits.

The changes to `docker/entrypoint.sh` are not related to running Piped as unprivilged user.

Apart from that, the changes generally look good to me and can be merged when split into different commits.

bjw-s:
> Please split the commit into two commits.
> 
> The changes to `docker/entrypoint.sh` are not related to running Piped as unprivilged user.
> 
> Apart from that, the changes generally look good to me and can be merged when split into different commits.

Thanks for the feedback! I've split up the PR in to two separate commits

Bnyro (approved):
Thank you!

Issue #2240, opened 2023-03-20
Select between audio descriptions / dubs


Myned:
### Describe the feature

Setting to switch between audio descriptions and dubbed voiceovers like YouTube.
An entry in the Preferences page is optional, but welcome.

### Why would this be useful to add?

This would provide the ability to view the original content or use voiceovers for accessibility purposes.

### Concept(s)

YouTube's implementation:

![image](https://user-images.githubusercontent.com/22806146/226416410-e766888c-900c-4444-810c-43da4c8d3d3d.png)

### Additional context

This may be considered a bug as Piped currently only uses videos with audio descriptions when available, and original if only dubs are available. I would imagine the current default should be the original video, but the ability to select between them is preferable.

[Example with audio description](https://piped.video/watch?v=Kn56bMZ9OE8)
[Example with only dubs](https://piped.video/watch?v=_8W2LIfl5RE)

It also seems to be a relatively new feature as I have only seen it with Tom Scott's newest videos, and I cannot find an official reference to it anywhere.

### Acknowledgements

- [X] I have searched the existing issues and this is **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this form.

endim8:
+1 for this, it's a bit annoying having the audio descriptions

bonswouar:
A related feature request that could be very useful would be to have better global settings for audio language.
For example :
- Preferred audio type: Dubbed / Audio Description / Original
- Preferred audio language _(might be different from UI language)_, this could probably be disabled if "Original" was selected before

PR #3892, opened 2024-12-15
fix: wrong selection of default audio language

bonswouar:
It would always fallback to `en` as the loop was over array keys, not values.
_I don't know much about JS but this `for..of` works in Firefox. Apparently it is ES6 compliant, supported pretty much everywhere except IE11_

**Really needed now that many English tracks of non-English videos are AI translated _(because yt enabled it by default)_ 🤢**

Notes:

- Ideally it would be great to be able to detect AI translated tracks, or "original language", not sure if it's possible with current extractor _(or with yt in general)_ EDIT: Apparently it is, and it's already included in proxy's response, nice!
- And also have a separate setting for Audio language (right now it's the same as UI language, which isn't so great) 

EDIT: I've added this to a related open feature request https://github.com/TeamPiped/Piped/issues/2240#issuecomment-2543971520

### DIFF
--- a/src/components/VideoPlayer.vue
+++ b/src/components/VideoPlayer.vue
@@ -613,7 +613,7 @@ export default {
                         // Set the audio language
                         const prefLang = this.getPreferenceString("hl", "en").substr(0, 2);
                         var lang = "en";
-                        for (var l in player.getAudioLanguages()) {
+                        for (var l of player.getAudioLanguages()) {
                             if (l == prefLang) {
                                 lang = l;
                                 return;


src/components/VideoPlayer.vue
```diff
@@ -613,7 +613,7 @@ export default {
                         // Set the audio language
                         const prefLang = this.getPreferenceString("hl", "en").substr(0, 2);
                         var lang = "en";
-                        for (var l in player.getAudioLanguages()) {
+                        for (var l of player.getAudioLanguages()) {
```
Bnyro, line 616: We could actually simplify this more, i.e.
```suggestion
          const lang = player.getAudioLanguages().includes(prefLang) ? prefLang : "en";
```
but that can be done in a different commit (I'll push that shortly after merging your PR).

Bnyro (approved):
>  don't know much about JS but this for..of works in Firefox. Apparently it is ES6 compliant, supported pretty much everywhere except IE11
>
Should be fine, it's used in other places at Piped too.

Thanks for the PR!

PR #3983, opened 2025-02-12
feat: add "Theater mode"

### LINKED ISSUE #3969: Recommendations take up space even if minimized
### Official Instance

- [ ] The bug is reproducible on the [official hosted instance](http://piped.video/), or is API-related.

### Describe the bug

I have the feeling that was different until recently...

"Minimize Recommendations" leaves the running video at the same size and still takes up space on the side:
Recommendations showing:
Image

Recommendations hidden:
Image

This also happens when recommendations are always hidden in the settings.

### To Reproduce

1. Open a video
2. Minimize Recommendations
3. Empty Space

### Expected behavior

Video would increase to full with of the window / screen.

### Logs/Errors

_No response_

Brave v1.74.51 on Nobara Linux 31, Plasma Version 6.2.5 Wayland.

### Additional context

_No response_

ac615223s5:
fix #3969

### DIFF
--- a/src/components/ChaptersBar.vue
+++ b/src/components/ChaptersBar.vue
@@ -1,6 +1,6 @@
 <template>
     <!-- desktop view -->
-    <div v-if="!mobileLayout" class="max-h-75vh max-w-35vw min-h-64 flex-col overflow-y-scroll lt-lg:hidden">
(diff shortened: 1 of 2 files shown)


Bnyro (approved):
Thank you!

PR #4034, opened 2025-04-29
Make list styling consistent in README.md

efb4f5ff-1298-471a-8973-3d47447115dc:
When looking through the README i noticed this odd spacing in the Not Under Active Development list compared to the other lists.
The reason for this is that its being grouped by category in the code but that cant be reflected visually. 

firefox_8quGYRdDRk

So you can do 2 things:
- Split up the lists like its done here: https://github.com/TeamPiped/Piped/blob/master/README.md?plain=1#L138-L151
- Make one big list

I chose the latter one as i dont think its important here to categorize applications that arent under active development

Before:

l9TuScmORm

After:

firefox_8Ya9ueDbi0

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,14 +152,11 @@ Contributions in any other form are also welcomed.
 
 
 **Not Under Active Development**
-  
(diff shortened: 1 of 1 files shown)


Bnyro (approved):
Good catch, thank you!

Issue #4000, opened 2025-03-05
Failed with error code 3016. See logs for more info


TheCrimsonLady:
### Official Instance

- [x] The bug is reproducible on the [official hosted instance](http://piped.video/), or is API-related.

### Describe the bug

Hello,

I have an issue with my instance that has persisted for around two weeks by now and is also reproducible on the official instance.
I get an error saying the title. Unfortunately, I cant seem to find anything in the logs (looked at all container logs individually) and grepping all docker logs for 3016 gets me zero hits. 
The error only occurs on iOS, as of now iOS 18.3.1. I also reinstalled Piped according to documentation with no change to my problem (except now videos stop playing after one minute on desktop...)

I have no idea how to troubleshoot this since the logs aren't telling me anything, so please tell me what I should include apart from my redacted config.properties and docker compose

### To Reproduce

1. Go to official piped instance on iOS 18.3.1
2. pick any video
3. it only displays the error from the title and does not play a video

### Expected behavior

The video should play normally

### Browser, and OS with Version.

Safari and Firefox on iOS 18.3.1

### Additional context

[config-properties.txt](https://github.com/user-attachments/files/19085581/config-properties.txt)
[docker-compose.txt](https://github.com/user-attachments/files/19085544/docker-compose.txt)

haderp:
I have the same error occurring with the latest docker builds and in my research came across this closed issue from shaka-player, https://github.com/shaka-project/shaka-player/issues/3509, though it remains unsolved and closed.  Another project has a similar issue still open that also deals with streaming protocols and iOS Safari found here, https://github.com/servo/servo/issues/34174.  I don't know if they are indeed related to this issue as I've not found anything definitive, the fact that one issue affects shaka-player and the other relates to HLS leads me to suspect it may be the same issue that is occurring here.  Sorry if this is a completely off base assumption, but thought I would share in the hopes it gives some idea to someone far more knowledgeable than myself.

haderp:
Sorry, I haven’t delved into that yet. Given that it doesn’t appear to be an issue being more widely reported I speculate it may be something environmentally specific that is the cause more so than something seen by iOS users en masse.  When I have time I planned to setup a test vm in order experiment with what could be the culprit.  I wish I had a better answer, hopefully someone with more knowledge of how the player works will chime in with a more educated idea on troubleshooting.

Insprill:
[...]

PR #4043, opened 2025-06-15
Add iOS MSE support and playsinline

Spendable1569:
iOS can support DASH playback via window.ManagedMediaSource.
This will fix https://github.com/TeamPiped/Piped/issues/4000, since it is caused by HLS problem.
playsinline prevents going to fullscreen with a black screen when in audio mode.


FireMasterK (approved):
Thanks, seems to be iOS 17+.

PR #4130, opened 2025-11-12
feat: add PWA standalone display mode to hide address bar

jollySleeper:
This PR adds the missing 'display: standalone' property to the PWA manifest configuration in vite.config.js. This change ensures that when the Piped frontend is installed as a PWA, it launches in standalone mode without showing the browser address bar, providing a more native app-like experience.

## Changes Made:
- Added `display: "standalone"` to PWA manifest for native-like appearance
- Added `description` property for better PWA metadata  
- Added `start_url: "/"` to specify app launch URL

## Testing:
To test these changes:
1. Build the application with `pnpm build`
2. Install the PWA from the browser
3. Launch the installed PWA - address bar should be hidden

This fix resolves the issue where the PWA would show browser chrome instead of running in full-screen standalone mode.

### DIFF
--- a/.github/workflows/docker-build-pr.yml
+++ b/.github/workflows/docker-build-pr.yml
@@ -1,0 +1,61 @@
+name: Docker Build PR
+
+on:
+  pull_request:
+    paths-ignore:
+      - "**.md"
+    branches:
+      - "**"
+
+jobs:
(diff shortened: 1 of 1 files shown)


jollySleeper:
This is not required.

PR #4234, opened 2026-05-09
feat: add local personalized trending feed

quantumvoid0:
This PR includes a new feature for the piped:trending page which allows users to enable a more personalized trending section from watch history.

It reads watch history from IndexedDB, ranks channels by watch count, fetches recent 5 videos from top 10 channels, Also added getPreferenceBoolean import. This feature can be turned on/off from preferences and also have the ability to show just the personalized section without global trending.

P.S:
Variables that are never reassigned should be const. so changed let->const... No functional difference, just cleaner code.

added RSS guard, without it, the RSS link would render with a URL like /feed/unauthenticated/rss?channels= (empty channels string) when not logged in and no subscriptions exist. Clicking it would hit the API with an empty parameter, likely returning an error or garbage. The null return + v-if just hides the button entirely in that case rather than showing a broken link.

## Summary by CodeRabbit

* **New Features**
  * Personalized trending on the Trending page using your watch history.
  * Two new preference toggles: enable personalized trending and optionally show only personalized recommendations (visible when watch history is enabled).
  * New localized labels and a “no watch history” message for personalized trending.

* **Bug Fixes**
  * RSS/feed icon now only appears when an RSS link is available.
  * Improved list update behavior to prevent inconsistent video lists.

[Review Change Stack](https://app.coderabbit.ai/change-stack/TeamPiped/Piped/pull/4234)


quantumvoid0:
I see there are some issues in the code as reported by coderabbit...ill see to fixing it first

src/components/TrendingPage.vue
```diff
@@ -18,42 +21,179 @@ import { useI18n } from "vue-i18n";
...
+
+const channelCache = new Map();
+const CHANNEL_CACHE_TTL = 5 * 60 * 1000;
+
+async function fetchChannelVideos(channelIds, historyCount) {
+    const cacheKey = channelIds.slice().sort().join(",") + "@" + historyCount;
+    const cached = channelCache.get(cacheKey);
+    if (cached && Date.now() - cached.ts < CHANNEL_CACHE_TTL) return cached.data;
+
+    const results = await Promise.allSettled(channelIds.map(id => fetchJson(apiUrl() + "/channel/" + id)));
```
FireMasterK: Can you use the unauthenticated channel feeds api, since this is quite expensive on the server? You can pass multiple channel IDs to it too.

quantumvoid0: surely, ill implement the fix..but there seems to be another issue preventing the code in this PR from running, about a week ago i noticed that videos section in channels no longer load, and this affected the trending page too as it reads from there, i dont think its just my side, bcz the piped.video server also had the same issue.

quantumvoid0: Update: the issue seem to be fixed now, ill update this PR with the requested changes soon

UltiomDesign:
I’d suggest adding a feature like this since the current trending page doesn’t really work for me.

One possible improvement would be introducing a self-hosted recommendation system. For example, users could run a local AI model on their own server or hardware (like an RTX 5090 Ti). The system could analyze the last ~100 watched videos using their titles, descriptions, and keywords, then periodically receive a small batch of new videos (e.g. 10 per hour) and rank or filter them based on user preferences.

YouTube’s recommendation system is one of its strongest features, and without a comparable alternative it’s difficult to fully switch away.

quantumvoid0:
@UltiomDesign right now, it doesnt really do what u think it does, it makes the feed a bit more personalized, what it does right now is fetches last watched youtube channels and a number of latest videos from them and display them, if you watch a channel more than another, that channel will get more priority..

but your idea of local ai model is nice, but that would require a custom ai model to be made as idk any right now which can provide such functions, and we will need it to work on low-end devices too to make it accessible to everyone

UltiomDesign:
@quantumvoid0 The idea of a local model is not necessarely really hard if we rely on text-based llm instead of a custom algorythm

Idea example:
On setup:  User inputs some usefull infos about what he likes to find (ex. "I am a huge fan of boxing and fighting video games")


- Piped backend opens the let's say the last 5 viewed videos that had at least 70% watch time -> Check the 15 first recommended videos on each
Take 10 Snapshot from the video & ask a model to find keywords (to generate keywords with confidence) and result
   ~ Example keywords: car, fire, explosion, money, fighting, boxing glothes, ring
   ~ Title: This video is about this and that
   ~ Description: The description says this video is about xxxxxxxxxxxx
   ~ Comments: Many comments enjoyed it, found it verry funny / Many users don't appreciate what was said at 0:45

Prompt a different (or new separate prompt) model with asking to rate the probability that the user will like this video (0.0 to 1.0)

Based of the result, we can make recommendations to the user that are fresh, unseen, and likely match way more content.

For models, there are plenty of local models that can be powerful enough to be run on 8GB of ram if not even less.
This is obvioulsy not VITAL as you are already working on Personalized Trending Feed, but I am sure that you would be interested by doing it.

